Issue #7, opened 2023-05-31
Event indexing


DavidMinarsch:
It might be worthwhile adding `sender` (the `msg.sender` of the matching Request) as an argument to `Deliver` event. This way, we would be able to index on `sender` off-chain and have event handlers that listen specifically for the events relating to a particular sender.

PR #10, opened 2023-10-19
Updating globals and adding argument in the event

kupermind:
- Updating globals;
- Verifying contracts;
- Adding argument to the Deliver event as specified in #7 .


77ph (approved):

contracts/AgentMech.sol
```diff
@@ -26,7 +26,7 @@ error NotEnoughPaid(uint256 provided, uint256 expected);
 /// @dev A Mech that is operated by the holder of an ERC721 non-fungible token.
 contract AgentMech is ERC721Mech {
     event Perform(address indexed sender, bytes32 taskHash);
-    event Deliver(uint256 requestId, bytes data);
+    event Deliver(address indexed sender, uint256 requestId, bytes data);
```
DavidMinarsch, line 29: Deliver is always the mech though

PR #11, opened 2023-10-23
feat: tracking undelivered request Ids

kupermind:
- Tracking of undelivered request Ids.


contracts/AgentMech.sol
```diff
@@ -22,19 +22,26 @@ error AgentNotFound(uint256 agentId);
...
 error NotEnoughPaid(uint256 provided, uint256 expected);
 
+/// @dev Request Id not found.
+/// @param requestId Request Id.
+error RequestIdNotFound(uint256 requestId);
+
 /// @title AgentMech - Smart contract for extending ERC721Mech
 /// @dev A Mech that is operated by the holder of an ERC721 non-fungible token.
 contract AgentMech is ERC721Mech {
-    event Perform(address indexed sender, bytes32 taskHash);
```
kupermind, line 28: This event is not used.

contracts/AgentMech.sol
```diff
@@ -22,19 +22,26 @@ error AgentNotFound(uint256 agentId);
...
 
     // Minimum required price
     uint256 public price;
+    // Number of undelivered requests
+    uint256 public numUndeliveredRequests;
 
     // Map of requests counts for corresponding addresses
     mapping (address => uint256) public mapRequestsCounts;
+    // Map of request Ids
+    mapping (uint256 => uint256[2]) public mapRequestIds;
```
kupermind, line 44: A two-way circular map stores all the undelivered requests and correctly removes elements when requests are delivered.

contracts/AgentMech.sol
```diff
@@ -95,4 +139,23 @@ contract AgentMech is ERC721Mech {
     function getRequestsCount(address account) external view returns (uint256 requestsCount) {
         requestsCount = mapRequestsCounts[account];
     }
+
+    /// @dev Gets the set of undelivered request Ids.
+    /// @return requestIds Set of undelivered request Ids.
+    function getUndeliveredRequestIds() external view returns (uint256[] memory requestIds) {
```
0xArdi, line 145: Do we need to worry about the number of undelivered requests being too large. Although unlikely, I think it may happen. Could that result in a big request to the rpc which would get dropped constantly?
I am thinking maybe it makes sense to either have this method take a maxNumReqs, or have another method that supports that, just to avoid this case.

DavidMinarsch: Good point. How about we slightly alter the method to allow to (optionally) set a batch size. Then you can get constant sized returns

0xArdi: @DavidMinarsch that should do

kupermind: I've accounted for both size and offset, please take a look.

contracts/AgentMech.sol
```diff
@@ -95,4 +144,42 @@ contract AgentMech is ERC721Mech {
...
+
+        // If size is zero, return all the requests
+        if (size == 0) {
+            size = numRequests;
+        }
+
+        // Check for the size + offset overflow
+        if (size + offset > numRequests) {
+            revert Overflow(size + offset, numRequests);
+        }
```
kupermind, line 164: We need both size and offset to access any number of requests.

contracts/AgentMech.sol
```diff
@@ -95,4 +144,42 @@ contract AgentMech is ERC721Mech {
...
+        if (size > 0) {
+            requestIds = new uint256[](size);
+
+            // The first request Id is the next request Id of the zero element in the request Ids map
+            uint256 curRequestId = mapRequestIds[0][1];
+            // Traverse requests a specified offset
+            for (uint256 i = 0; i < offset; ++i) {
+                // Next request Id of the current element based on the current request Id
+                curRequestId = mapRequestIds[curRequestId][1];
+            }
```
kupermind, line 175: Skip the initial offset number of requests.

0xArdi (approved):

PR #20, opened 2023-12-05
Nevermined

0xArdi:
TBD


contracts/integrations/nevermined/AgentMechSubscription.sol
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        // Check for the number of credits available in the subscription
+        uint256 creditsBalance = IERC1155(subscriptionNFT).balanceOf(msg.sender, subscriptionTokenId);
+        uint256 creditsPerRequest = price;
+        if (creditsBalance < creditsPerRequest) {
+            revert NotEnoughCredits(creditsBalance, creditsPerRequest);
+        }
+
+        _locked = 1;
```
0xArdi, line 100: I think we need to account for pending requests. 
As it is right now, a user can send as many request as they want assuming they have balance for at least one of them, and there is no delivery made. 
This has to be handled, either on or off-chain. I think it would be easier to handle on-chain, but we could also do it off-chain.

What do you think @kupermind @DavidMinarsch ?

kupermind: Makes sense.

contracts/integrations/nevermined/AgentMechSubscription.sol
```diff
@@ -0,0 +1,143 @@
...
+
+        // Burn credits of the request Id sender upon delivery
+        IERC1155(subscriptionNFT).burn(mapRequestAddresses[requestId], subscriptionTokenId, price);
+
+        // Return the request data
+        requestData = data;
+
+        _locked = 1;
+    }
+
```
DavidMinarsch, line 123: What I'm confused about here is that now the price is still static. The idea was that requests can be dynamically priced...

contracts/integrations/nevermined/AgentFactorySubscription.sol
```diff
@@ -0,0 +1,64 @@
...
+    /// @param agentHash IPFS CID hash of the agent metadata.
+    /// @param creditsPerRequest Number of credits to pay for request via subscription.
+    /// @param subscriptionNFT Subscription address.
+    /// @param subscriptionTokenId Subscription token Id.
+    /// @return agentId The id of a created agent.
+    /// @return mech The created mech instance address.
+    function create(
+        address agentOwner,
+        bytes32 agentHash,
+        uint256 creditsPerRequest,
```
DavidMinarsch, line 47: Can these ever be updated? The idea was to have dynamic pricing per request...

kupermind: The price can be updated by the Mech's `setPrice()` function

contracts/integrations/nevermined/AgentMechSubscription.sol
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        _locked = 1;
+    }
+
+    /// @dev Performs actions before the delivery of a request.
+    /// @param requestId Request Id.
+    /// @param data Self-descriptive opaque data-blob.
+    /// @return requestData Data for the request processing.
+    function _preDeliver(uint256 requestId, bytes memory data) internal override returns (bytes memory requestData) {
```
0xArdi, line 108: should take a `cost` argument. Which should be used in the burn below. This will require a new deliver method.

contracts/integrations/nevermined/AgentMechSubscription.sol
```diff
@@ -0,0 +1,143 @@
...
+        // Reentrancy guard
+        if (_locked > 1) {
+            revert ReentrancyGuard();
+        }
+        _locked = 2;
+
+        // Burn credits of the request Id sender upon delivery
+        IERC1155(subscriptionNFT).burn(mapRequestAddresses[requestId], subscriptionTokenId, price);
```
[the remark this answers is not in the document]
0xArdi: should use the argument mentioned above

contracts/integrations/nevermined/AgentMechSubscription.sol
kupermind: This will be extracted from the data itself

scripts/deployment/deploy_02_agent_factory.js
```diff
@@ -43,7 +43,7 @@ async function main() {
 
     // Transaction signing and execution
     console.log("2. EOA to deploy AgentFactory pointed to AgentRegistry");
-    const AgentFactory = await ethers.getContractFactory("AgentFactory");
```
DavidMinarsch, line 46: Make this configurable for both approaches

scripts/deployment/verify_04_agent_mech.js
```diff
@@ -6,9 +6,13 @@ const parsedData = JSON.parse(dataFromJSON);
 const agentRegistryAddress = parsedData.agentRegistryAddress;
 const agentId = parsedData.agentId;
```
DavidMinarsch, line 7: same

PR #21, opened 2023-12-21
refactor: adding credit check for all pending requests

kupermind:
- Adding credit check for all pending request Ids.

### DIFF
--- a/contracts/integrations/nevermined/AgentMechSubscription.sol
+++ b/contracts/integrations/nevermined/AgentMechSubscription.sol
@@ -90,11 +90,11 @@ contract AgentMechSubscription is AgentMech {
             revert NoDepositAllowed(amount);
         }
 
-        // Check for the number of credits available in the subscription
+        // Check for the number of credits available in the subscription vs total number of credits needed
         uint256 creditsBalance = IERC1155(subscriptionNFT).balanceOf(msg.sender, subscriptionTokenId);
-        uint256 creditsPerRequest = price;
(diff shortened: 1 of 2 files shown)


contracts/integrations/nevermined/AgentMechSubscription.sol
```diff
@@ -90,11 +90,11 @@ contract AgentMechSubscription is AgentMech {
...
         }
 
-        // Check for the number of credits available in the subscription
+        // Check for the number of credits available in the subscription vs total number of credits needed
         uint256 creditsBalance = IERC1155(subscriptionNFT).balanceOf(msg.sender, subscriptionTokenId);
-        uint256 creditsPerRequest = price;
-        if (creditsBalance < creditsPerRequest) {
-            revert NotEnoughCredits(creditsBalance, creditsPerRequest);
+        uint256 creditsPerPendingRequests = mapUndeliveredRequestsCounts[msg.sender] * price;
+        if (creditsBalance < creditsPerPendingRequests) {
```
0xArdi, line 96: Shouldnt this be less than or equal? Since this is a preReq check? If the credits balance is same as the number of credits to be used by pending reqs, we should revert, no?

kupermind: Oh that must be just made with +1, as we have not accounted for the incoming request. After that it's a strict `<`, as if we have enough credits to cover previous pending and a current request, we are fine.

0xArdi (approved):

PR #48, opened 2024-12-12
feat: first part of service registry replacing agent registry

DavidMinarsch:


contracts/AgentFactory.sol
```diff
@@ -12,45 +12,71 @@ interface IAgentRegistry {
     function create(address agentOwner, bytes32 agentHash) external returns (uint256 agentId);
 }
 
+
+// Service Registry interface
```
DavidMinarsch, line 16: Lots of duplication now - needs cleaning

contracts/AgentFactory.sol
```diff
@@ -12,45 +12,71 @@ interface IAgentRegistry {
...
 /// @title Agent Factory - Periphery smart contract for managing agent and mech creation
 contract AgentFactory is GenericManager {
     event CreateMech(address indexed mech, uint256 indexed agentId, uint256 indexed price);
 
     // Agent factory version number
     string public constant VERSION = "1.1.0";
 
-    // Agent registry address
-    address public immutable agentRegistry;
+    // Service registry address
```
DavidMinarsch, line 47: Factory should point to service registry. IMO factory could even be deprecated and we add this function to marketplace itself...

contracts/AgentFactory.sol
```diff
@@ -12,45 +12,71 @@ interface IAgentRegistry {
...
             revert Paused();
         }
+        address multisig;
+        (, multisig, , , , , ) = IService(serviceRegistry).mapServices(serviceId);
+        // below not needed, checked in OlasMech constructor
+        // if (state != IService.ServiceState.Deployed) {
+        //     revert WrongServiceState(uint256(state), serviceId);
+        // }
+        bytes32 salt = keccak256(abi.encode(multisig, serviceId));
+        // multisig is isOperator() for the mech
```
DavidMinarsch, line 77: This requires us to fly our own mech implementation. Otherwise we have an unnecessary layer of NFT in between

contracts/AgentMech.sol
```diff
@@ -76,9 +103,57 @@ error Overflow(uint256 provided, uint256 max);
...
-/// @title AgentMech - Smart contract for extending ERC721Mech
+/// @dev Wrong state of a service.
+/// @param state Service state.
+/// @param serviceId Service Id.
+error WrongServiceState(uint256 state, uint256 serviceId);
+
+/**
+ * @dev A Mech that is operated by the multisig of an Olas service
+ */
+contract OlasMech is Mech, ImmutableStorage {
```
DavidMinarsch, line 114: This is almost identical to ERC721Mech except the constructor and isOperator implementations

contracts/AgentMech.sol
```diff
@@ -76,9 +103,57 @@ error Overflow(uint256 provided, uint256 max);
...
+    /// @param _token Address of the registry contract
+    /// @param _tokenId The token ID
+    constructor(address _token, uint256 _tokenId) {
+        bytes memory initParams = abi.encode(_token, _tokenId);
+        address multisig;
+        IService.ServiceState state;
+        (, multisig, , , , , state) = IService(_token).mapServices(_tokenId);
+        if (state != IService.ServiceState.Deployed) {
+            revert WrongServiceState(uint256(state), _tokenId);
+        }
```
DavidMinarsch, line 124: We check the service exists and is live

contracts/AgentMech.sol
```diff
@@ -76,9 +103,57 @@ error Overflow(uint256 provided, uint256 max);
...
+    }
+
+    function isOperator(address signer) public view override returns (bool) {
+        (address _token, uint256 _tokenId) = abi.decode(
+            readImmutable(),
+            (address, uint256)
+        );
+        address multisig;
+        (, multisig, , , , , ) = IService(_token).mapServices(_tokenId);
+        return multisig == signer;
```
DavidMinarsch, line 150: operator simplified - we now check if the multisig is the signer

contracts/AgentMech.sol
```diff
@@ -401,6 +478,7 @@ contract AgentMech is ERC721Mech {
         _locked = 1;
     }
 
+    /// TODO: We need the same system to work with Nevermined pricing.
```
DavidMinarsch, line 481: Important that we consider this right from the start. Ideally the factory takes care of creating both types of mechs

contracts/AgentFactory.sol
kupermind: I'd vote for that as well

PR #51, opened 2024-12-13
refactor: mech manager to control marketplaces and factories

kupermind:
- Mech manager to control marketplaces and factories;
- Payment handling for request deliveries;
- Everything is able to be processed by mech marketplaces only.


contracts/MechManager.sol
```diff
@@ -0,0 +1,279 @@
...
+        // Record factory that created agent mech
+        mapAgentMechFactories[mech] = mechFactory;
+
+        emit CreateMech(mech, serviceId);
+    }
+
+    /// @dev Sets mech marketplace statues.
+    /// @param mechMarketplaces Mech marketplace contract addresses.
+    /// @param statuses Corresponding whitelisting statues.
+    function setMechMarketplaceStatuses(address[] memory mechMarketplaces, bool[] memory statuses) external {
```
DavidMinarsch, line 158: Why do we have multiple marketplaces? That's really not needed and just adding confusion for a user of the system

contracts/MechMarketplace.sol
```diff
@@ -58,19 +57,25 @@ contract MechMarketplace is IErrorsMarketplace {
     // Maximum response time
     uint256 public immutable maxResponseTimeout;
     // Mech karma contract address
-    address public immutable karmaProxy;
```
DavidMinarsch, line 61: why changed the name?

contracts/MechMarketplace.sol
```diff
@@ -58,19 +57,25 @@ contract MechMarketplace is IErrorsMarketplace {
...
+    uint256 public collectedFees;
     // Number of undelivered requests
     uint256 public numUndeliveredRequests;
     // Number of total requests
     uint256 public numTotalRequests;
     // Reentrancy lock
     uint256 internal _locked = 1;
 
+    // Map of request payments
+    mapping(uint256 => uint256) public mapRequestIdPayments;
```
DavidMinarsch, line 78: Why do we need this? Can we not get away with a fee counter per mech?

contracts/MechMarketplace.sol
```diff
@@ -141,19 +150,37 @@ contract MechMarketplace is IErrorsMarketplace {
...
 
-        // This should never be the case
-        if (mech == address(0)) {
-            revert ZeroAddress();
+        // Get request Id payment
+        uint256 payment = mapRequestIdPayments[requestId];
+
+        // Process payment
+        if (payment > 0) {
+            uint256 fee = IMechManager(mechManager).fee();
```
DavidMinarsch, line 168: I don't want to have multiple marketplaces. Don't like the complexity this introduces. imo it's much simpler to scope fees logic to mechs. Then marketplace interface stays the same and simply the factory supports new types of mechs. A mech specifies its fee model that gets retrieved here.

contracts/MechMarketplace.sol
```diff
@@ -141,19 +150,37 @@ contract MechMarketplace is IErrorsMarketplace {
...
+            payment = payment - (payment * fee) / 100;
+
+            // Transfer payment
+            (bool success, ) = deliveryMech.call{value: msg.value}("");
```
DavidMinarsch, line 172: Unnecessary gas overhead no? why not make it pull rather than push. I.e. just have a mapping that's mech specific that contains their accrued fees.

contracts/MechMarketplace.sol
```diff
@@ -233,7 +260,7 @@ contract MechMarketplace is IErrorsMarketplace {
         mechDelivery.requester = msg.sender;
```
DavidMinarsch, line 260: why not store the payment here instead of the new mapping?
mechDelivery.payment = msg.value;

contracts/MechMarketplace.sol
```diff
@@ -242,8 +269,18 @@ contract MechMarketplace is IErrorsMarketplace {
         // Increase the total number of requests
         numTotalRequests++;
 
+        // Must never happen
+        if (mapRequestIdPayments[requestId] > 0) {
+            revert RequestPaid(requestId);
+        }
+        // Record request payment
+        mapRequestIdPayments[requestId] = msg.value;
```
DavidMinarsch, line 277: see above

contracts/MechMarketplace.sol
```diff
@@ -242,8 +269,18 @@ contract MechMarketplace is IErrorsMarketplace {
...
+
+        // Record request counts in global sets
+        IMechManager(mechManager).increaseRequestsCounts(msg.sender);
```
DavidMinarsch, line 280: why do we need this now?

contracts/MechMarketplace.sol
```diff
@@ -321,16 +358,29 @@ contract MechMarketplace is IErrorsMarketplace {
         // Increase the amount of requester delivered requests
         mapDeliveryCounts[requester]++;
         // Increase the amount of mech service multisig delivered requests
-        mapMechServiceDeliveryCounts[mechService]++;
+        mapMechServiceDeliveryCounts[mechServiceMultisig]++;
+
+        // Record request counts in global sets
+        IMechManager(mechManager).increaseDeliveryCounts(requester, msg.sender, mechServiceMultisig);
```
DavidMinarsch, line 364: for what?

contracts/MechMarketplace.sol
```diff
@@ -321,16 +358,29 @@ contract MechMarketplace is IErrorsMarketplace {
...
 
         _locked = 1;
     }
 
+    // TODO: send to BBB, wrap first or manage on BBB side?
+    function drain() external {
+        if (collectedFees == 0) {
+            revert ZeroValue();
+        }
+    }
```
DavidMinarsch, line 382: BBB knows nothing about Marketplace and only accepts OLAS or native (wrapped). So we should send the wrapped version.

contracts/MechMarketplace.sol
```diff
@@ -353,6 +403,7 @@ contract MechMarketplace is IErrorsMarketplace {
                 getDomainSeparator(),
                 keccak256(
                     abi.encode(
+                        block.timestamp,
```
DavidMinarsch, line 406: why the change?

contracts/MechManager.sol
```diff
@@ -0,0 +1,279 @@
...
+
+/// @dev Wrong length of two arrays.
+/// @param numValues1 Number of values in a first array.
+/// @param numValues2 Number of values in a second array.
+error WrongArrayLength(uint256 numValues1, uint256 numValues2);
+
+/// @title MechManager - Smart contract for mech manager
+/// @author Aleksandr Kuperman - <aleksandr.kuperman@valory.xyz>
+/// @author Andrey Lebedev - <andrey.lebedev@valory.xyz>
+contract MechManager {
```
DavidMinarsch, line 38: I don't see the benefit of this.

The mech marketplace could have a simple create method:
```
    /// @dev Registers service as a mech.
    /// @param serviceId Service id.
    /// @param mechFactory Mech factory address.
    /// @return mech The created mech instance address.
    function create(uint256 serviceId, address mechFactory, bytes memory payload) external returns (address mech) {
        // Check for factory status
        if (!mapMechFactories[mechFactory]) {
            revert UnauthorizedAccount(mechFactory);
        }

        mech = IMechFactory(mechFactory).createMech(address(this), serviceRegistry, serviceId, payload);

        // This should never be the case
        if (mech == address(0)) {
            revert ZeroAddress();
        }

        // Record factory that created agent mech
        mapAgentMechFactories[mech] = mechFactory;

        emit CreateMech(mech, serviceId);
    }
```

contracts/MechManager.sol
```diff
@@ -0,0 +1,279 @@
...
+            revert AlreadyInitialized();
+        }
+
+        owner = msg.sender;
+        fee = _fee;
+    }
+
+    /// @dev Changes the mechManager implementation contract address.
+    /// @param newImplementation New implementation contract address.
+    function changeImplementation(address newImplementation) external {
```
DavidMinarsch, line 96: This implies governance, so might as well make the marketplace itself a proxy.

contracts/MechManager.sol
```diff
@@ -0,0 +1,279 @@
...
+
+        owner = newOwner;
+        emit OwnerUpdated(newOwner);
+    }
+
+    /// @dev Registers service as a mech.
+    /// @param serviceId Service id.
+    /// @param mechFactory Mech factory address.
+    /// @return mech The created mech instance address.
+    function create(uint256 serviceId, address mechFactory, bytes memory payload) external returns (address mech) {
```
DavidMinarsch, line 136: see above

contracts/MechManager.sol
```diff
@@ -0,0 +1,279 @@
...
+            mapMechMarketplaces[mechMarketplaces[i]] = statuses[i];
+        }
+
+        emit SetMechMarketplaceStatuses(mechMarketplaces, statuses);
+    }
+
+    /// @dev Sets mech factory statues.
+    /// @param mechFactories Mech marketplace contract addresses.
+    /// @param statuses Corresponding whitelisting statues.
+    function setMechFactoryStatuses(address[] memory mechFactories, bool[] memory statuses) external {
```
DavidMinarsch, line 183: It's enough to have a boolean status flag for the mech factories imo

contracts/MechManager.sol
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        // Decrease the number of undelivered requests
+        numUndeliveredRequests--;
+        // Increase the amount of requester delivered requests
+        mapDeliveryCounts[requester]++;
+        // Increase the amount of agent mech delivery counts
+        mapAgentMechDeliveryCounts[agentMech]++;
+        // Increase the amount of mech service multisig delivered requests
+        mapMechServiceDeliveryCounts[mechServiceMultisig]++;
```
DavidMinarsch, line 232: better on marketplace. just confusing having this manager

contracts/MechManager.sol
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /// @dev Validates agent mech.
+    /// @param agentMech Agent mech address.
+    /// @return status True, if the mech is valid.
+    function checkMechValidity(address agentMech) external view returns (bool status) {
+        // TODO: shall we also check the status of the factory?
+        // TODO: if yes, what if the factory was de-whitelisted? all the mechs then become invalid
+        status = mapAgentMechFactories[agentMech] != address(0);
+    }
```
DavidMinarsch, line 242: We do the same as in stkaing, once created its valid. Only check at creation time.

contracts/MechManager.sol
```diff
@@ -0,0 +1,279 @@
...
+
+        mech = IMechFactory(mechFactory).createMech(address(this), serviceRegistry, serviceId, payload);
+
+        // This should never be the case
+        if (mech == address(0)) {
+            revert ZeroAddress();
+        }
+
+        // Record factory that created agent mech
+        mapAgentMechFactories[mech] = mechFactory;
```
DavidMinarsch, line 150: Not needed imo.

But where are we recording the mechs that the marketplace has registered on it?

Currently there's no on-chain mapping. That makes discovery very hard.

contracts/Karma.sol
```diff
@@ -37,6 +37,7 @@ contract Karma {
     // Contract owner
     address public owner;
 
+    // TODO This must be fetched from mech manager and be removed / inaccessible from here
```
DavidMinarsch, line 40: Why?

contracts/AgentMech.sol
```diff
@@ -327,11 +332,13 @@ contract AgentMech is OlasMech {
 
     /// @dev Delivers a request by a marketplace.
     /// @notice This function ultimately calls mech marketplace contract to finalize the delivery.
+    /// @param mechMarketplace Mech marketplace address.
     /// @param requestId Request id.
     /// @param data Self-descriptive opaque data-blob.
     /// @param mechStakingInstance Mech staking instance address (optional).
     /// @param mechServiceId Mech operator service Id.
     function deliverToMarketplace(
+        address mechMarketplace,
```
DavidMinarsch, line 341: Multiple marketplaces is not helpful

contracts/MechMarketplace.sol
kupermind: Cleaner - nobody cares in the contract if it's proxy or not, the setup matters

contracts/MechMarketplace.sol
kupermind: I don't understand what is meant by that. Funds don't go to mechs before requests are delivered, but payment is taken during the request. What is your suggestion?

contracts/MechMarketplace.sol
kupermind: Refactoring... but fee needs further clarification

contracts/MechMarketplace.sol
kupermind: Obsolete

contracts/MechMarketplace.sol
kupermind: to add more randomness into numbers

contracts/MechManager.sol
kupermind: Now moved in the marketplace

contracts/MechManager.sol
kupermind: Then anybody is able to create a factory with its mech and then deliver the request, so at least we ned to check created mechs.

contracts/MechManager.sol
kupermind: See above

contracts/Karma.sol
kupermind: Irrelevant now, stays as is

contracts/MechMarketplace.sol
kupermind: Implemented

contracts/MechManager.sol
kupermind: Implemented

contracts/MechMarketplace.sol
kupermind: oh great, of course, so easy :O

DavidMinarsch (approved):

PR #53, opened 2024-12-16
refactor: majority of mechs logic go into OlasMech

kupermind:
- Majority of mechs logic go into OlasMech


contracts/MechMarketplace.sol
```diff
@@ -196,6 +206,29 @@ contract MechMarketplace is IErrorsMarketplace {
         );
     }
 
+    /// @dev Calculates payment for request delivery.
+    /// @param mech Delivery mech address.
+    /// @param payment Payment amount.
+    function _calculatePayment(
```
kupermind, line 212: Maybe we should move this to the side of a factory or a mech itself? Just make sure mech is not able to manipulate. This would allow to proceed with a default scenario.

contracts/MechMarketplace.sol
```diff
@@ -301,8 +371,9 @@ contract MechMarketplace is IErrorsMarketplace {
         emit SetMechFactoryStatuses(mechFactories, statuses);
     }
 
+    // TODO: leave optional fields or remove?
```
DavidMinarsch, line 374: tbd

contracts/MechMarketplace.sol
```diff
@@ -415,17 +486,9 @@ contract MechMarketplace is IErrorsMarketplace {
...
+        // Check delivery mech and get its service multisig
         address mechServiceMultisig = checkMech(msg.sender, deliveryMechStakingInstance, deliveryMechServiceId);
 
-        // Get the staked service info for the mech
-        IStaking.ServiceInfo memory serviceInfo =
-            IStaking(deliveryMechStakingInstance).getServiceInfo(deliveryMechServiceId);
-        // Check that staked service multisig is the priority mech mech service
-        if (!IMech(msg.sender).isOperator(serviceInfo.multisig)) {
-            revert UnauthorizedAccount(msg.sender);
-        }
```
DavidMinarsch, line 427: why removed?

kupermind: Exactly same check is performed in a function above: `checkMech()`

contracts/MechMarketplace.sol
```diff
@@ -196,6 +206,29 @@ contract MechMarketplace is IErrorsMarketplace {
         );
     }
 
+    /// @dev Calculates payment for request delivery.
+    /// @param mech Delivery mech address.
+    /// @param payment Payment amount.
+    function _calculatePayment(
+        address mech,
+        uint256 payment
+    ) internal virtual returns (uint256 mechPayment, uint256 marketplaceFee) {
```
Sfgangloff, line 215: In the function _calculatePayment we should also add the marketplaceFee to collectedFee, no ? 
Otherwise drain() does nothing it seems.

kupermind: Good catch!

PR #55, opened 2024-12-18
V0.4.0 internal audit

77ph:
- Internal audit against refactor branch


audit/internal2/README.md
```diff
@@ -0,0 +1,204 @@
...
+Several checks are obtained automatically. They are commented. <br>
+All automatic warnings are listed in the following file, concerns of which we address in more detail below: <br>
+[slither-full](https://github.com/valory-xyz/autonolas-staking-programmes/blob/main/audits/internal2/analysis/slither_full.txt) 
+Most of the issues raised by instrumental analysis are outside the scope of the audit. <br>
+
+
+### Issue
+#### Critical? _preDeliver() passed low then price and zero balance 
+```
+        // Check for the number of credits available in the subscription
```
Sfgangloff, line 24: _preDeliver() should probably be reverted whenever creditsToBurn < creditsBalance. 
This means in particular that when there is not enough credit in the balance, the mech can't deliver (I think it would not want to). Probably need to change the Karma mechanism in case using subscription actually (cause the mech does not deliver because of low balance, not its fault).

kupermind (approved):

PR #56, opened 2024-12-18
refactor: split delivery rates according to mech type

kupermind:
- Split delivery rates according to mech type.


contracts/MechFixedPrice.sol
```diff
@@ -5,54 +5,19 @@ import {OlasMech} from "./OlasMech.sol";
...
-    }
-
-    /// @dev Performs actions before the request is posted.
-    /// @param amount Amount of payment in wei.
-    function _preRequest(uint256 amount, uint256, bytes memory) internal virtual override {
-        // Check the request payment
-        if (amount < price) {
-            revert NotEnoughPaid(amount, price);
-        }
-    }
```
kupermind, line 37: This all becomes obsolete, as `price` becomes `maxDeliveryRate` and is now uniformal across mechs.

The checks are done on MM for now, but ultimately will be taken out to a corresponding separate escrow contract

contracts/MechMarketplace.sol
```diff
@@ -177,30 +175,33 @@ contract MechMarketplace is IErrorsMarketplace {
...
     }
 
-    /// @dev Calculates payment for request delivery.
+    /// @dev Calculates payment and fee based on delivery rates and mech type.
     /// @param mech Delivery mech address.
-    /// @param payment Payment amount.
+    /// @param deliveryRate Delivery rate.
+    /// @param mechPayment Mech payment.
+    /// @param marketplaceFee Marketplace fee.
     function _calculatePayment(
```
kupermind, line 183: This becomes a single calculation based on deliveryRate. Clearly can be adjusted further in its final escrow function as it is virtual.

contracts/MechMarketplace.sol
```diff
@@ -365,6 +364,29 @@ contract MechMarketplace is IErrorsMarketplace {
         emit SetMechFactoryStatuses(mechFactories, statuses);
     }
 
+    // Check and escrow delivery rate
+    function _checkAndEscrowDeliveryRate(address mech) internal {
```
kupermind, line 368: We hold whatever incoming funds on MM (escrow) relevant contracts based on a `maxDeliveryRate`.

contracts/MechMarketplace.sol
```diff
@@ -554,11 +579,18 @@ contract MechMarketplace is IErrorsMarketplace {
             revert UnauthorizedAccount(msg.sender);
         }
 
-        // Transfer balance
-        // solhint-disable-next-line avoid-low-level-calls
-        (bool success, ) = msg.sender.call{value: balance}("");
-        if (!success) {
-            revert TransferFailed(address(0), address(this), msg.sender, balance);
+        // Get mech type
+        IMech.MechType mechType = IMech(msg.sender).mechType();
```
kupermind, line 583: The payout is also managed depending on a mech type.

contracts/OlasMech.sol
```diff
@@ -221,6 +239,18 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
         emit Deliver(msg.sender, requestId, requestData);
     }
 
+    /// @dev Sets the new max delivery rate.
+    /// @param newMaxDeliveryRate New max delivery rate.
+    function changeMaxDeliveryRate(uint256 newMaxDeliveryRate) external virtual onlyOperator {
```
kupermind, line 244: Same function for `maxDeliveryRate` update for every mech.

contracts/MechMarketplace.sol
```diff
@@ -177,30 +175,33 @@ contract MechMarketplace is IErrorsMarketplace {
...
-                revert ZeroValue();
-            }
+        IMech.MechType mechType = IMech(mech).mechType();
+        uint256 maxDeliveryRate = IMech(mech).maxDeliveryRate();
 
-            // Record payment into mech balance
-            mapMechBalances[mech] += mechPayment;
+        // TODO what if fee is zero just because the delivery rate is in the order of 1..10_000?
+        // Calculate mech payment and marketplace fee
+        marketplaceFee = (deliveryRate * fee) / 10_000;
```
77ph, line 192: Not sure.
marketplaceFee = (deliveryRate * fee) / 10_000;
Something is logically wrong here with the dimensions or the name.
marketplaceFee [token?] = Rate [?] * 10 / 10_000
for 
mechPayment = deliveryRate - marketplaceFee; We need the same dimensions for all quantities.

77ph (approved):

contracts/MechMarketplace.sol
kupermind: Fee is in `10_000` form, but TDB

contracts/MechMarketplace.sol
```diff
@@ -177,30 +175,33 @@ contract MechMarketplace is IErrorsMarketplace {
...
-            // Process payment
-            marketplaceFee = (payment * fee) / 10_000;
-            mechPayment = payment - marketplaceFee;
-
-            // Check for zero value
-            if (mechPayment == 0) {
-                revert ZeroValue();
-            }
+        IMech.MechType mechType = IMech(mech).mechType();
+        uint256 maxDeliveryRate = IMech(mech).maxDeliveryRate();
```
Sfgangloff, line 188: Seems maxDeliveryRate is not used in this function

Sfgangloff (approved):

contracts/MechMarketplace.sol
```diff
@@ -177,30 +175,40 @@ contract MechMarketplace is IErrorsMarketplace {
...
-            // Process payment
-            marketplaceFee = (payment * fee) / 10_000;
-            mechPayment = payment - marketplaceFee;
-
-            // Check for zero value
-            if (mechPayment == 0) {
-                revert ZeroValue();
-            }
+        // Get actual delivery rate
+        uint256 actualDeliveryRate = IMech(mech).getFinalizedDeliveryRate(requestId);
```
kupermind, line 190: I believe this is a required change - get the finalized delivery rate that is recorded for every mech.

contracts/MechMarketplace.sol
```diff
@@ -365,6 +371,29 @@ contract MechMarketplace is IErrorsMarketplace {
...
+            if (msg.value < maxDeliveryRate) {
+                revert InsufficientBalance(msg.value, maxDeliveryRate);
+            }
+        } else {
+            // Check that there is no incoming deposit
+            if (msg.value > 0) {
+                revert NoDepositAllowed(msg.value);
+            }
+
+            // TODO Add onERC1155Received to the receiving contract
```
DavidMinarsch, line 391: What for?

contracts/MechMarketplace.sol
DavidMinarsch: The problem with this approach is the rounding. For micropayments on Nevermined you'll have marketplaceFee round down to 0.

Alternative approach is to simply implement the fee at the withdrawal stage. It's 10x simpler. Simply when a mech withdraws their balance the marketplace takerate gets applied uniformly.

contracts/integrations/nevermined/MechNeverminedSubscription.sol
```diff
@@ -118,52 +68,45 @@ contract MechNeverminedSubscription is OlasMech {
...
-        }
+        mapRequestIdFinalizedRates[requestId] = deliveryRate;
 
-        emit DeliverPrice(requestId, deliverPrice, creditsToBurn);
+        emit DeliveryRateFinalized(requestId, deliveryRate, creditsToBurn);
 
         _locked = 1;
     }
 
+    // TODO This migrates to its corresponding escrow contract
```
DavidMinarsch, line 82: what escrow contract?

DavidMinarsch: You mean the marketplace's?

kupermind:
@DavidMinarsch @Sfgangloff 

> if mech.is_native_withdrawal() {
> }
> 
> if mech.is_erc20_withdrawal() {
> }

There's no way this is going to be pretty - this means all these interfaces. We still have to introduce something like `payment_type` value, which makes it a mech type, even if you don't call it so.

Re: ERC-20 - MM still needs to escrow it, there's just no other way. You can't account for allowance() as by the time of delivery the funds could be already gone, so only transferFrom, and thus - escrow. So what do we have? Escrow for pure `msg.value`, escrow for ERC-20, then why keep pushing for MM having all this burden?

For the Nevermined stuff we discussed that the fee must be managed as pre-paid, and thus all the escrow is going to do is to just check if there's enough credits on a subscription - and transferFrom from subscription and burn on withdraw.

Why having these endless if-else-s in the MM itself when it's super complicated already?

re: `marketplaceFee` - that's why there's a `TODO` comment there, as it's not set in stones. Things with subscriptions are more clear now. That is why we need to take all the payment management stuff outside of MM, otherwise we'll keep on redeploying it!

contracts/MechMarketplace.sol
```diff
@@ -365,6 +371,29 @@ contract MechMarketplace is IErrorsMarketplace {
...
+            // Get max subscription rate for escrow from sender
+            IERC1155(subscriptionNFT).safeTransferFrom(msg.sender, address(this), subscriptionTokenId, maxDeliveryRate, "");
```
Sfgangloff, line 393: Where is the subscriptionNFT/subscriptionTokenId defined ?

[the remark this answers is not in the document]
Sfgangloff: > Why does this happen here? should happen on delivery

Seems that this is transferring to the marketplace and then on delivery it is transfered to the balance of the mech who delivers. But true that it probably payment could probably be transfered directly from the requester to the mech's balance without having the marketplace as an intermediate holder.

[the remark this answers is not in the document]
kupermind: We don't need to transfer anything, just check, because the subscription is already paid, although I'd prefer to transfer credits to MM because by the time the payment needs to be done in bulk - there might not be credits left, and this is the issue.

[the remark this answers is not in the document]
kupermind: Let the micropayments and microcalculations be managed by MM processors, and the fee is applied when the mech pulls with a single transfer to that mech.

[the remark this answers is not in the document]
kupermind: This is about any external tokens - unless taken into hold - there's no guarantee the payment is going to be there by the checkout time.

contracts/integrations/nevermined/BalanceTrackerSubscription.sol
```diff
@@ -0,0 +1,216 @@
...
+    // Check and record delivery rate
+    function checkAndRecordDeliveryRate(
+        address mech,
+        address requester,
+        uint256,
+        bytes memory
+    ) external payable {
+        // Check for marketplace access
+        if (msg.sender != mechMarketplace) {
+            revert ManagerOnly(msg.sender, mechMarketplace);
```
DavidMinarsch, line 99: Call it `MarketplaceOnly`

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,332 @@
...
+        if (paymentData.length == 32) {
+            // Extract token address
+            token = abi.decode(paymentData, (address));
+            if (token != address(0) && msg.value > 0) {
+                revert NoDepositAllowed(msg.value);
+            }
+        } else if (paymentData.length > 0) {
+            revert InvalidPayloadLength(paymentData.length, 32);
+        }
+
```
DavidMinarsch, line 137: There's no handling of msg.value below in case of native token transfer

I never approved giving up on one-off payments per mech requests as we currently support. The new system should support the same way mechs worked in the past as well as new types..

you can trivially apply msg.value to the balance logic below when it is positive and matches the native token type for the balancertracker

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,332 @@
...
+            actualDeliveryRate = maxDeliveryRate;
+        }
+
+        // Record payment into mech balance
+        mapMechBalances[mech][token] += actualDeliveryRate;
+
+        emit MechPaymentCalculated(mech, requestId, actualDeliveryRate, rateDiff);
+    }
+
+    // TODO buyBackBurner does not account for other tokens but WETH, OLAS
```
DavidMinarsch, line 191: We should for fixed price only use the native token or OLAS imo, so this is not an issue

if in the future other tokens are needed we cross that bridge then

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,332 @@
...
+    function drain(address token) external {
+        // Reentrancy guard
+        if (_locked > 1) {
+            revert ReentrancyGuard();
+        }
+        _locked = 2;
+
+        uint256 localCollectedFees = mapCollectedFees[token];
+
+        // TODO Limits
```
DavidMinarsch, line 202: No problem if it gets sent often, so no limit needed

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,332 @@
...
+    function processPayment(address token) external returns (uint256 mechPayment, uint256 marketplaceFee) {
+        // Reentrancy guard
+        if (_locked > 1) {
+            revert ReentrancyGuard();
+        }
+        _locked = 2;
+
+        // Get mech balance
+        uint256 balance = mapMechBalances[msg.sender][token];
+        // TODO limits?
```
DavidMinarsch, line 251: We could check round up always, meaning incentives are to withdraw later.

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,332 @@
...
+        if (balance == 0) {
+            revert ZeroValue();
+        }
+
+        // Calculate mech payment and marketplace fee
+        uint256 fee = IMechMarketplace(mechMarketplace).fee();
+        marketplaceFee = (balance * fee) / 10_000;
```
DavidMinarsch, line 258: See above, suggest rounding up so we always incentivise later, rather than earlier withdrawals

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,332 @@
...
+        _withdraw(token, balance);
+
+        emit Withdraw(msg.sender, token, balance);
+
+        _locked = 1;
+    }
+
+    // Deposits token funds for requester.
+    function deposit(address token, uint256 amount) external {
+        // TODO Accept deposits from mechs as well?
```
DavidMinarsch, line 308: As mentioned above, whilst this is a nice additional feature, the default of one-off payments also needs to be retained for fixed price!

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,332 @@
...
+
+        if (token == address(0)) {
+            revert ZeroAddress();
+        }
+
+        // TODO: safe transfer?
```
DavidMinarsch, line 314: IMO you should have one FixedPriceNative that's scoped to native token and one FixedPriceERC20 that's scoped to erc20, not mix implementations

DavidMinarsch (commented):
we must ensure a non-priority mech can only deliver if their own maxDeliveryRate is below or equal that of the priority mech

DavidMinarsch:
It's safe to remove the staking related stuff no? The activity checking would still function afaik

DavidMinarsch:
Should let mech allow to call itself - marketplace will collect fees and could be an interesting use case at best, at worst nothing changes

contracts/BalanceTrackerFixedPrice.sol
kupermind: Agree, much easier to deal with

contracts/MechMarketplace.sol
kupermind: Outside of the scope

PR #57, opened 2024-12-18
refactor: separate escrows for each of mech type

kupermind:
- Separate escrows for each of mech type.

### DIFF
--- a/contracts/EscrowBase.sol
+++ b/contracts/EscrowBase.sol
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+/// @dev Provided zero marketplace address.
+error ZeroMarketplaceAddress();
+
+/// @dev Only `manager` has a privilege, but the `sender` was provided.
+/// @param sender Sender address.
+/// @param manager Required sender address as a manager.
+error ManagerOnly(address sender, address manager);
+
+abstract contract EscrowBase {
+    event Withdraw(address indexed mech, uint256 amount);
+    event Drained(uint256 collectedFees);
+
+    // Mech marketplace address
+    address public immutable mechMarketplace;
+
+    // Collected fees
+    uint256 public collectedFees;
+    // Reentrancy lock
+    uint256 internal _locked = 1;
+
+    // Map of mech => its current balance
+    mapping(address => uint256) public mapMechBalances;
+
+    constructor(address _mechMarketplace) {
+        // Check for zero address
+        if (_mechMarketplace == address(0)) {
+            revert ZeroMarketplaceAddress();
+        }
+
+        mechMarketplace = _mechMarketplace;
+    }
+
+    // Check and escrow delivery rate
+    function checkAndEscrowDeliveryRate(address mech) external virtual payable;
+
+    /// @dev Drains collected fees by sending them to a Buy back burner contract.
+    function drain() external virtual;
+
+    function adjustBalances(address mech, uint256 mechPayment, uint256 marketplaceFee) external virtual {
+        if (msg.sender != mechMarketplace) {
+            revert ManagerOnly(msg.sender, mechMarketplace);
+        }
+
+        // Record payment into mech balance
+        mapMechBalances[mech] += mechPayment;
+
+        // Record collected fee
+        collectedFees += marketplaceFee;
+    }
+
+    /// @dev Withdraws funds for a specific mech.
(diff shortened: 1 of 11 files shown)


contracts/EscrowBase.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: MIT
```
kupermind, line 1: Base contract for escrows. Main idea: each mech type has its own respective escrow contract to verify payment rates and manage fees.

contracts/MechFactoryFixedPrice.sol
```diff
@@ -10,7 +10,7 @@ error IncorrectDataLength(uint256 provided, uint256 expected);
 
 /// @title Mech Factory Basic - Periphery smart contract for managing basic mech creation
 contract MechFactoryFixedPrice {
-    event CreateBasicMech(address indexed mech, uint256 indexed serviceId, uint256 indexed price);
+    event CreateFixedPriceMech(address indexed mech, uint256 indexed serviceId, uint256 maxDeliveryRate);
```
kupermind, line 13: `maxDeliveryRate` sounds nicer than `price`

contracts/MechMarketplace.sol
```diff
@@ -194,21 +169,26 @@ contract MechMarketplace is IErrorsMarketplace {
...
 
-        // Record payment into mech balance
-        mapMechBalances[mech] += mechPayment;
-
-        // Record collected fee
-        collectedFees += marketplaceFee;
+        // Record mech payment and marketplace fee into corresponding escrow balances
+        IMech.MechType mechType = IMech(mech).mechType();
+        address escrow = mapMechTypeEscrows[mechType];
+        IEscrow(escrow).adjustBalances(mech, mechPayment, marketplaceFee);
```
kupermind, line 191: This logic should probably also go to the escrow, where we just pass the MM current fee, since balance adjustment also happens on that side.

contracts/MechMarketplace.sol
```diff
@@ -371,27 +347,30 @@ contract MechMarketplace is IErrorsMarketplace {
...
     }
 
-    // Check and escrow delivery rate
-    function _checkAndEscrowDeliveryRate(address mech) internal {
-        IMech.MechType mechType = IMech(mech).mechType();
-        uint256 maxDeliveryRate = IMech(mech).maxDeliveryRate();
+    /// @dev Sets mech type escrows.
+    /// @param mechTypes Mech types.
+    /// @param escrows Corresponding escrow addresses.
+    function setMechTypeEscrows(IMech.MechType[] memory mechTypes, address[] memory escrows) external {
```
kupermind, line 353: Each mech type has its own escrow contract to manage type-specific payments / fees.

contracts/EscrowBase.sol
DavidMinarsch: The concept of the escrow means we have unnecessary transfers from Nevermined to here and then from here to withdrawal destination.

contracts/MechMarketplace.sol
```diff
@@ -194,21 +169,26 @@ contract MechMarketplace is IErrorsMarketplace {
             revert ZeroValue();
         }
 
+        // TODO What to do if there is rateDIff leftovers? Keep as a fee or record into sender's map?
+        uint256 rateDiff;
+        if (actualDeliveryRate > deliveryRate) {
+            rateDiff = actualDeliveryRate - deliveryRate;
+        }
```
DavidMinarsch, line 176: in case of fixed price it just gets consumed, otherwise all we care for is lower or equal than max (because it won't be spent anyway)

contracts/MechMarketplace.sol
```diff
@@ -194,21 +169,26 @@ contract MechMarketplace is IErrorsMarketplace {
...
+
         // TODO what if fee is zero just because the delivery rate is in the order of 1..10_000?
         // Calculate mech payment and marketplace fee
-        marketplaceFee = (deliveryRate * fee) / 10_000;
-        mechPayment = deliveryRate - marketplaceFee;
+        marketplaceFee = (actualDeliveryRate * fee) / 10_000;
+        mechPayment = actualDeliveryRate - marketplaceFee;
```
DavidMinarsch, line 181: DO NOT apply this here. rounding issues will render it unusable. Can only apply take rate at withdrawal for large (lower bound enforced!) amounts

DavidMinarsch (approved):

PR #58, opened 2024-12-19
refactor: avoid micropayments

kupermind:
- Avoid micropayments.


contracts/EscrowFixedPrice.sol
```diff
@@ -66,65 +76,121 @@ contract EscrowFixedPrice is EscrowBase {
...
-        emit Withdraw(msg.sender, balance);
+        emit Withdraw(msg.sender, token, balance);
 
         _locked = 1;
     }
 
-    /// @dev Drains collected fees by sending them to a Buy back burner contract.
-    function drain() external virtual override {
+    /// @dev Withdraws funds for a specific account.
+    function withdrawAccount(address token) external virtual override {
```
Sfgangloff, line 177: So this function allows to withdraw from a token and the limit of amount of this withdrawal is the balance of the requester in the marketplace (so if all is used for requests nothing can be withdrawn using this function). But how is it ensured that the requester can't withdraw from this token by other means than through this the marketplace and this function ?

kupermind: This is outdated now :)

kupermind: Balances are checked - you can't withdraw somehow from another place.

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,327 @@
...
+            token = abi.decode(paymentData, (address));
+            if (token != address(0) && msg.value > 0) {
+                revert NoDepositAllowed(msg.value);
+            }
+        } else if (paymentData.length > 0) {
+            revert InvalidPayloadLength(paymentData.length, 32);
+        }
+
+        // Get account balance
+        uint256 balance = mapRequesterBalances[requester][token];
```
kupermind, line 139: Get requester balance - it must be sufficient to cover request.

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,327 @@
...
+        if (balance < maxDeliveryRate) {
+            revert InsufficientBalance(balance, maxDeliveryRate);
+        }
+
+        // Record request token
+        mapRequestIdTokens[requestId] = token;
+
+        // Adjust account balance
+        balance -= maxDeliveryRate;
+        mapRequesterBalances[requester][token] = balance;
```
kupermind, line 151: Requester balance is adjusted with the mech max delivery rate

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,327 @@
...
+    /// @param requestId Request Id.
+    /// @param deliveryRate Requested delivery rate.
+    function finalizeDeliveryRate(address mech, address requester, uint256 requestId, uint256 deliveryRate) external {
+        // Check for marketplace access
+        if (msg.sender != mechMarketplace) {
+            revert ManagerOnly(msg.sender, mechMarketplace);
+        }
+
+        // Get actual delivery rate
+        uint256 actualDeliveryRate = IMech(mech).getFinalizedDeliveryRate(requestId);
```
kupermind, line 166: Gets the actual rate from the mech for the request delivery

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,329 @@
...
+        address token = mapRequestIdTokens[requestId];
+
+        uint256 rateDiff;
+        if (maxDeliveryRate > actualDeliveryRate) {
+            // Return back requester overpayment debit
+            rateDiff = maxDeliveryRate - actualDeliveryRate;
+            mapRequesterBalances[requester][token] += rateDiff;
+        } else {
+            actualDeliveryRate = maxDeliveryRate;
+        }
```
kupermind, line 183: Now the delivery rate is exactly adjusted with wither the leftovers, or max being a cap.

contracts/MechMarketplace.sol
```diff
@@ -43,10 +43,10 @@ contract MechMarketplace is IErrorsMarketplace {
     event ImplementationUpdated(address indexed implementation);
     event MarketplaceParamsUpdated(uint256 fee, uint256 minResponseTimeout, uint256 maxResponseTimeout);
     event SetMechFactoryStatuses(address[] mechFactories, bool[] statuses);
-    event SetMechTypeEscrows(IMech.MechType[] mechTypes, address[] escrows);
+    event SetPaymentTypeBalanceTrackers(uint8[] paymentTypes, address[] balanceTrackers);
```
kupermind, line 46: `uint8` is a better solution, this way no need to modify MechMarketplace when another mech type is added.

contracts/MechMarketplace.sol
```diff
@@ -150,47 +150,6 @@ contract MechMarketplace is IErrorsMarketplace {
         );
     }
 
-    /// @dev Calculates payment and fee based on delivery rates and mech type.
-    /// @param mech Delivery mech address.
-    /// @param requestId Request Id.
-    /// @param deliveryRate Delivery rate.
-    /// @param mechPayment Mech payment.
-    /// @param marketplaceFee Marketplace fee.
-    function _calculatePayment(
```
kupermind, line 159: This is migrated to each balance tracker accordingly, and modified as per above.

contracts/BalanceTrackerFixedPrice.sol
```diff
@@ -0,0 +1,332 @@
...
+        uint256 balance = mapMechBalances[msg.sender][token];
+        // TODO limits?
+        if (balance == 0) {
+            revert ZeroValue();
+        }
+
+        // Calculate mech payment and marketplace fee
+        uint256 fee = IMechMarketplace(mechMarketplace).fee();
+        marketplaceFee = (balance * fee) / 10_000;
+        mechPayment = balance - marketplaceFee;
```
kupermind, line 259: Calculate fee on withdraw only

contracts/integrations/nevermined/BalanceTrackerSubscription.sol
```diff
@@ -0,0 +1,200 @@
...
+    // Subscription NFT
+    address public immutable subscriptionNFT;
+    // Subscription token Id
+    uint256 public immutable subscriptionTokenId;
+
+    // Reentrancy lock
+    uint256 internal _locked = 1;
+
+    // Map of requester => current credit balance
+    mapping(address => uint256) public mapRequesterBalances;
```
kupermind, line 63: Note in order to calculate precise credits for requesters - we record credits, not debits, as credits are already bought in subscription.

contracts/integrations/nevermined/BalanceTrackerSubscription.sol
```diff
@@ -0,0 +1,200 @@
...
+        // Get requester actual subscription balance
+        uint256 subscriptionBalance = IERC1155(subscriptionNFT).balanceOf(requester, subscriptionTokenId);
+
+        // Adjust requester balance with maxDeliveryRate credits
+        balance += maxDeliveryRate;
+
+        // Check the request delivery rate for a fixed price
+        if (subscriptionBalance < balance) {
+            revert InsufficientBalance(subscriptionBalance, balance);
+        }
```
kupermind, line 116: Note the difference - we figure out the actual subscription amount and how much the next request costs. The overall value must not be bigger than the balance in credits already accounted for.

contracts/integrations/nevermined/BalanceTrackerSubscription.sol
```diff
@@ -0,0 +1,216 @@
...
+            if (rateDiff > balance) {
+                revert Overflow(rateDiff, balance);
+            }
+
+            // Adjust requester balance
+            balance -= rateDiff;
+            mapRequesterBalances[requester] = balance;
+        } else {
+            actualDeliveryRate = maxDeliveryRate;
+        }
```
kupermind, line 170: Here we can finally exactly account for actual delivery rate and reduce credits from requester, if delivery rate cost less.

contracts/integrations/nevermined/BalanceTrackerSubscription.sol
```diff
@@ -0,0 +1,216 @@
...
+
+        // Record payment into mech balance
+        mapMechBalances[mech] += actualDeliveryRate;
```
kupermind, line 173: Not sure in the current model if we need to track that at all. Even in the previous version mechs never took any credits directly. Need to sync with Nvm again.

contracts/integrations/nevermined/BalanceTrackerSubscription.sol
```diff
@@ -0,0 +1,216 @@
...
+
+        emit MechPaymentCalculated(mech, requestId, actualDeliveryRate, rateDiff);
+
+        _locked = 1;
+    }
+
+    /// @dev Processes payment.
+    function processPayment(address requester) external {
```
kupermind, line 181: The logic is exactly the same as it was before on Nvm subscription, but before it was burning credits for the requester per each delivery. Now as we are waiting for credits to accumulate to burn in bigger amount, it's unlikely that requester is going to call the function to trigger its credit burning themselves. That's why we need to provide the address explicitly.

contracts/integrations/nevermined/BalanceTrackerSubscription.sol
```diff
@@ -0,0 +1,216 @@
...
+        // TODO limits and correct balance value
+        if (creditsToBurn == 0) {
+            revert InsufficientBalance(0, 0);
+        }
+
+        // Clear balances
+        mapRequesterBalances[requester] = 0;
+
+        // Burn credits of the request Id sender upon delivery
+        IERC1155(subscriptionNFT).burn(requester, subscriptionTokenId, creditsToBurn);
```
kupermind, line 210: Exact logic as before.

kupermind: Compare with deleted one below.

contracts/integrations/nevermined/MechNeverminedSubscription.sol
```diff
@@ -42,50 +42,25 @@ contract MechNeverminedSubscription is OlasMech {
...
-        uint256 creditsBalance = IERC1155(subscriptionNFT).balanceOf(account, subscriptionTokenId);
-
-        // Adjust the amount of credits to burn if the deliver price is bigger than the amount of credits available
-        if (creditsToBurn > creditsBalance) {
-            creditsToBurn = creditsBalance;
-        }
-
-        // Burn credits of the request Id sender upon delivery
-        if (creditsToBurn > 0) {
-            IERC1155(subscriptionNFT).burn(account, subscriptionTokenId, creditsToBurn);
```
kupermind, line 83: This is the original logic - burn credits per request.

contracts/MechMarketplace.sol
```diff
@@ -409,6 +411,7 @@ contract MechMarketplace is IErrorsMarketplace {
             revert UnauthorizedAccount(priorityMechStakingInstance);
         }
 
+        // TODO Shall we allow other mechs to post requests to mechs? Or completely prohibit mechs to post requests?
```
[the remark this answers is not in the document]
DavidMinarsch: You're not solving manipulation (sybil) by preventing self requests and you might make other nice future usecases impossible by an arbitrary restiction. Unless we have a really good reason to restrict, we should not.

DavidMinarsch (approved):

PR #59, opened 2024-12-20
refactor: payments as subscription and as micropayments for requests

kupermind:
- Payments as subscription and as micropayments for requests


contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -0,0 +1,241 @@
...
+error InvalidPayloadLength(uint256 provided, uint256 expected);
+
+/// @dev Failure of a transfer.
+/// @param token Address of a token.
+/// @param from Address `from`.
+/// @param to Address `to`.
+/// @param amount Amount value.
+error TransferFailed(address token, address from, address to, uint256 amount);
+
+abstract contract BalanceTrackerFixedPriceBase {
```
kupermind, line 49: The common ground of everything fixed price

contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -0,0 +1,242 @@
...
+        address requester,
+        uint256
+    ) external payable {
+        // Check for marketplace access
+        if (msg.sender != mechMarketplace) {
+            revert MarketplaceOnly(msg.sender, mechMarketplace);
+        }
+
+        // Check for native value
+        _checkNativeValue();
```
kupermind, line 101: This simply checks if working with a native token or another one.

contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -0,0 +1,242 @@
...
+        // Get account balance
+        uint256 balance = mapRequesterBalances[requester];
+
+        // Check the request delivery rate for a fixed price
+        if (balance < maxDeliveryRate) {
+            // Get balance difference
+            uint256 balanceDiff = maxDeliveryRate - balance;
+            // Adjust balance
+            balance += _getRequiredFunds(requester, balanceDiff);
+        }
```
kupermind, line 115: This is key - the balance can be top-upped via receive / deposit function. However if there's a lack of it - charge "at place", that makes the behavior exact equivalent to what there currently is.

contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -0,0 +1,242 @@
...
+        balance -= maxDeliveryRate;
+        mapRequesterBalances[requester] = balance;
+    }
+
+    /// @dev Finalizes mech delivery rate based on requested and actual ones.
+    /// @param mech Delivery mech address.
+    /// @param requester Requester address.
+    /// @param requestId Request Id.
+    /// @param maxDeliveryRate Requested max delivery rate.
+    function finalizeDeliveryRate(address mech, address requester, uint256 requestId, uint256 maxDeliveryRate) external {
```
kupermind, line 127: This is unchanged from previous version - except for the token field removal.

contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -0,0 +1,242 @@
...
+        // Record payment into mech balance
+        mapMechBalances[mech] += actualDeliveryRate;
+
+        emit MechPaymentCalculated(mech, requestId, actualDeliveryRate, rateDiff);
+    }
+
+    function _drain(uint256 amount) internal virtual;
+
+    /// @dev Drains collected fees by sending them to a Buy back burner contract.
+    function drain() external {
```
kupermind, line 159: Unchanged.

contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -0,0 +1,242 @@
...
+            revert ReentrancyGuard();
+        }
+        _locked = 2;
+
+        // Get mech balance
+        uint256 balance = mapMechBalances[msg.sender];
+        // TODO minimal balance value to account for the round-off
+        if (balance == 0 || balance < 10_000) {
+            revert InsufficientBalance(balance, 10_000);
+        }
```
kupermind, line 196: Need to define minimal value here to compare against.

kupermind: The rest is unchanged.

contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -0,0 +1,242 @@
...
+        mapMechBalances[msg.sender] = 0;
+
+        // Process withdraw
+        _withdraw(balance);
+
+        _locked = 1;
+    }
+
+    /// @dev Withdraws funds for a specific requester account.
+    function withdraw() external {
```
kupermind, line 221: Unchanged, just removed the token field

contracts/BalanceTrackerFixedPriceNative.sol
```diff
@@ -0,0 +1,79 @@
...
+        if (_wrappedNativeToken == address(0)) {
+            revert ZeroAddress();
+        }
+
+        wrappedNativeToken = _wrappedNativeToken;
+    }
+
+    function _checkNativeValue() internal virtual override {}
+
+    function _getRequiredFunds(address, uint256 balanceDiff) internal virtual override returns (uint256) {
```
kupermind, line 40: Check for sufficient `msg.value` to top-up

contracts/BalanceTrackerFixedPriceToken.sol
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    function _checkNativeValue() internal virtual override {
+        // Check for msg.value
+        if (msg.value > 0) {
+            revert NoDepositAllowed(msg.value);
+        }
+    }
+
+    function _getRequiredFunds(address requester, uint256 balanceDiff) internal virtual override returns (uint256) {
```
kupermind, line 48: Get top-up tokens

kupermind: probably needs to compare bool, right @77ph ?

contracts/MechFactoryFixedPriceNative.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import {MechFixedPriceNative} from "./MechFixedPriceNative.sol";
```
kupermind, line 4: Just the same as for the native token, no logic change

contracts/MechFixedPriceToken.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import {OlasMech} from "./OlasMech.sol";
+
+/// @title MechFixedPriceToken - Smart contract for OlasMech that accepts a fixed price payment for services in native token.
+contract MechFixedPriceToken is OlasMech {
```
kupermind, line 7: Same as a native token, just the payment type is different

contracts/MechMarketplace.sol
```diff
@@ -351,8 +350,7 @@ contract MechMarketplace is IErrorsMarketplace {
         uint256 priorityMechServiceId,
         address requesterStakingInstance,
         uint256 requesterServiceId,
-        uint256 responseTimeout,
-        bytes memory paymentData
```
kupermind, line 355: No need fir payment data, if anything additional is needed for the future - can be stored under `requestId` in the mech itself.

contracts/OlasMech.sol
```diff
@@ -15,8 +15,9 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
     event RevokeRequest(address indexed sender, uint256 requestId);
 
     enum PaymentType {
-        FixedPrice,
-        Subscription
+        FixedPriceNative,
+        FixedPriceToken,
+        NvmSubscription
```
kupermind, line 20: Giving better naming

contracts/OlasMech.sol
```diff
@@ -214,6 +213,11 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
             return requestData;
         }
 
+        // Check for max delivery rate compared to requested one
+        if (maxDeliveryRate > mechDelivery.deliveryRate) {
+            revert Overflow(maxDeliveryRate, mechDelivery.deliveryRate);
+        }
```
kupermind, line 219: This is important

contracts/MechMarketplace.sol
DavidMinarsch: Hm. I'd just leave a default `data` field.

contracts/BalanceTrackerFixedPriceBase.sol
DavidMinarsch: My idea was to do the rounding up (ceil, not floor). This way, we don't need to define a hardcode on the balance as the marketplaceFee is always strictly positive (e.g. 1% fee, 1 cent withdrawn -> fee is 1 cent; 50 cent withdrawn -> fee is 1 cent, 1 dollar withdrawn -> fee is 1 cent

contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -0,0 +1,242 @@
...
+        _checkNativeValue();
+
+        // Get mech max delivery rate
+        uint256 maxDeliveryRate = IMech(mech).maxDeliveryRate();
+
+        // Get account balance
+        uint256 balance = mapRequesterBalances[requester];
+
+        // Check the request delivery rate for a fixed price
+        if (balance < maxDeliveryRate) {
```
DavidMinarsch, line 110: hm in the case of native value we don't account for native value being transferred in in all cases. this is a bug

kupermind: Good one

PR #61, opened 2024-12-22
chore: partial changes

DavidMinarsch:


contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -106,6 +106,12 @@ abstract contract BalanceTrackerFixedPriceBase {
         // Get account balance
         uint256 balance = mapRequesterBalances[requester];
 
+        // Update balance with native value
```
DavidMinarsch, line 109: The native value was not handled correctly

contracts/BalanceTrackerFixedPriceNative.sol
```diff
@@ -37,11 +37,8 @@ contract BalanceTrackerFixedPriceNative is BalanceTrackerFixedPriceBase {
 
     function _checkNativeValue() internal virtual override {}
 
-    function _getRequiredFunds(address, uint256 balanceDiff) internal virtual override returns (uint256) {
-        if (msg.value < balanceDiff) {
-            revert InsufficientBalance(msg.value, balanceDiff);
-        }
-        return msg.value;
+    function _getRequiredFunds(address, uint256) internal virtual override returns (uint256) {
+        return 0;
```
DavidMinarsch, line 41: Now it matches the method as well (in the case of native token there is no get, just push)

contracts/MechFactoryFixedPriceNative.sol
```diff
@@ -27,6 +27,8 @@ contract MechFactoryFixedPriceNative {
         uint256 serviceId,
         bytes memory payload
     ) external returns (address mech) {
+        // TODO: restrict all factories to be called from marketplace only - makes it easier to monitor the system
```
DavidMinarsch, line 30: Important imo

contracts/MechMarketplace.sol
```diff
@@ -108,26 +108,23 @@ contract MechMarketplace is IErrorsMarketplace {
...
     mapping(address => uint256) public mapNonces;
-    // Set of mechs created by this marketplace
-    address[] public setMechs;
+    // Mapping of service ids to mechs
+    mapping(uint256 => address) public mapServiceIdMech;
 
 
     /// @dev MechMarketplace constructor.
     /// @param _serviceRegistry Service registry contract address.
-    /// @param _stakingFactory Staking factory contract address.
```
DavidMinarsch, line 117: Dropping all staking stuff. We just need the counters for the activity checkers, not these various variables imo

contracts/MechMarketplace.sol
```diff
@@ -275,8 +272,9 @@ contract MechMarketplace is IErrorsMarketplace {
 
         // Record factory that created a mech
         mapAgentMechFactories[mech] = mechFactory;
-        // Add mech address into the global set
-        setMechs.push(mech);
+        // Add mapping
+        mapServiceIdMech[serviceId] = mech;
```
DavidMinarsch, line 276: needed later

contracts/MechMarketplace.sol
```diff
@@ -332,24 +330,17 @@ contract MechMarketplace is IErrorsMarketplace {
...
+    /// @param requesterServiceId Requester service Id (optional).
     /// @param responseTimeout Relative response time in sec.
-    /// @param paymentData Additional payment-related request data, if applicable.
+    /// @param paymentData Additional payment-related request data (optional).
     /// @return requestId Request Id.
     function request(
         bytes memory data,
-        address priorityMech,
-        address priorityMechStakingInstance,
         uint256 priorityMechServiceId,
```
DavidMinarsch, line 343: passing in the serviceId should be enough as we can retrieve the mech from that

contracts/MechMarketplace.sol
```diff
@@ -463,12 +443,6 @@ contract MechMarketplace is IErrorsMarketplace {
             revert ZeroAddress();
         }
 
-        // Check that the delivery mech is not a requester
```
DavidMinarsch, line 466: As discussed, dropping

contracts/MechMarketplace.sol
```diff
@@ -437,13 +421,9 @@ contract MechMarketplace is IErrorsMarketplace {
     /// @notice This function can only be called by the mech delivering the request.
     /// @param requestId Request id.
     /// @param requestData Self-descriptive opaque data-blob.
-    /// @param deliveryMechStakingInstance Delivery mech staking instance address (optional).
-    /// @param deliveryMechServiceId Mech service Id.
```
DavidMinarsch, line 441: don't need staking. service id can be retrieved from mech

contracts/MechMarketplace.sol
```diff
@@ -548,69 +522,41 @@ contract MechMarketplace is IErrorsMarketplace {
...
             revert ZeroValue();
         }
 
+        uint256 mechServiceId = IMech(mech).tokenId();
+
         // Check mech validity as it must be created and recorded via this marketplace
-        if (mapAgentMechFactories[mech] == address(0)) {
+        if (mapServiceIdMech[mechServiceId] == mech) {
             revert UnauthorizedAccount(mech);
         }
```
DavidMinarsch, line 556: consistency check

contracts/MechMarketplace.sol
```diff
@@ -548,69 +522,41 @@ contract MechMarketplace is IErrorsMarketplace {
...
     }
 
     /// @dev Checks for mech validity.
     /// @dev mech Agent mech contract address.
-    /// @param mechStakingInstance Agent mech staking instance address.
     /// @param mechServiceId Agent mech service Id.
-    /// @return multisig Service multisig address.
+    /// @return multisig
     function checkMech(
```
DavidMinarsch, line 543: probably can further simplify this method, but for now its ok

contracts/MechMarketplace.sol
```diff
@@ -619,25 +565,21 @@ contract MechMarketplace is IErrorsMarketplace {
...
         address requester,
-        address requesterStakingInstance,
         uint256 requesterServiceId
     ) public view {
         // Check for requester service
         if (requesterServiceId > 0) {
-            address multisig = checkServiceAndGetMultisig(requesterStakingInstance, requesterServiceId);
+            address multisig = checkServiceAndGetMultisig(requesterServiceId);
 
             // Check staked service multisig
```
DavidMinarsch, line 579: need to update docstring to drop staking reference

contracts/OlasMech.sol
```diff
@@ -14,7 +14,7 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
     event Request(address indexed sender, uint256 requestId, bytes data);
     event RevokeRequest(address indexed sender, uint256 requestId);
 
-    enum PaymentType {
+    enum PaymentType { // this level of the contract should not have to specify its possible extensions. So Payment type needs to be something that's extensible, string or number..
```
DavidMinarsch, line 17: @kupermind please fix!

contracts/OlasMech.sol
```diff
@@ -38,8 +38,10 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
     // Mech marketplace address
     address public immutable mechMarketplace;
     // Mech payment type
+    // TODO - can it be id or string so its' forward compatible?
```
DavidMinarsch, line 41: see above

contracts/OlasMech.sol
```diff
@@ -51,6 +53,7 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
     // Reentrancy lock
     uint256 internal _locked = 1;
 
+    // TODO do we need the next two maps? Is it for staking activity checks?
```
DavidMinarsch, line 56: @kupermind we should check if we can remove either this map or the one on the marketplace. Feels like info is being duplicated

contracts/OlasMech.sol
```diff
@@ -71,9 +75,14 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
         address _mechMarketplace,
         address _serviceRegistry,
         uint256 _serviceId,
-        uint256 _maxDeliveryRate,
+        uint256 _maxDeliveryRate, // TODO - for subscriptions, why do we even specify this here; it should be specified by the requester
```
DavidMinarsch, line 78: To discuss

contracts/OlasMech.sol
```diff
@@ -138,12 +147,12 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
     /// @dev Registers a request.
     /// @param account Requester account address.
-    /// @param data Self-descriptive opaque data-blob.
     /// @param requestId Request Id.
+    /// @param data Self-descriptive opaque data-blob.
     function _request(
         address account,
-        bytes memory data,
-        uint256 requestId
+        uint256 requestId,
```
DavidMinarsch, line 154: order for match of other maths

contracts/OlasMech.sol
```diff
@@ -214,20 +223,18 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
         }
 
         // Check for max delivery rate compared to requested one
-        if (maxDeliveryRate > mechDelivery.deliveryRate) {
+        if (maxDeliveryRate < mechDelivery.deliveryRate) {
```
DavidMinarsch, line 226: typo?

contracts/OlasMech.sol
```diff
@@ -341,7 +344,13 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
             (address, uint256)
         );
 
-        (, address multisig, , , , , ) = IServiceRegistry(serviceRegistry).mapServices(serviceId);
+        (, address multisig, , , , , IServiceRegistry.ServiceState state) =
+            IServiceRegistry(_serviceRegistry).mapServices(tokenId());
+
+        // Check for correct service state
```
DavidMinarsch, line 350: Imo we should always check for the deploy state when the mech is used

contracts/MechMarketplace.sol
```diff
@@ -275,8 +272,9 @@ contract MechMarketplace is IErrorsMarketplace {
 
         // Record factory that created a mech
         mapAgentMechFactories[mech] = mechFactory;
-        // Add mech address into the global set
-        setMechs.push(mech);
```
kupermind, line 279: I would leave this in some form. For now there's no mech tracker, just events. `numMechs` then becomes meaningless.

contracts/OlasMech.sol
kupermind: No, this was correct. The delivery mech cannot have a bigger delivery rate than requester agreed for. `mechDelivery.deliveryRate` is the max delivery rate that requester chose when selecting the priority mech.

contracts/BalanceTrackerFixedPriceBase.sol
kupermind: I don't like this, `Base` should not know about `msg.value` or not. I will refactor further.

kupermind: There's a `_checkNativeValue()` function above, at least need to move it there.

contracts/MechMarketplace.sol
DavidMinarsch: We now have the mapping, don't also need a set. The numMechs is now tracked explicitly as a variable

contracts/OlasMech.sol
kupermind: Agree, I didn't like this hardcoding either

contracts/OlasMech.sol
kupermind: I believe you are correct here, seems like just the legacy from when mechs were created not with the MM interactions.

contracts/OlasMech.sol
kupermind: It's still the mech level - so mech must decide its max delivery rate.

contracts/OlasMech.sol
```diff
@@ -61,6 +64,7 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
     mapping(uint256 => uint256[2]) public mapRequestIds;
     // Map of request Id => sender address
     mapping(uint256 => address) public mapRequestAddresses;
+    // TODO is a global nonce not sufficient?
```
kupermind, line 67: This one is not used at all.

contracts/OlasMech.sol
kupermind: hash is pretty, as always :)

contracts/MechMarketplace.sol
kupermind: This naturally resolves itself

contracts/OlasMech.sol
kupermind: Resolved to hashes

contracts/OlasMech.sol
```diff
@@ -38,8 +38,10 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
     // Mech marketplace address
     address public immutable mechMarketplace;
     // Mech payment type
+    // TODO - can it be id or string so its' forward compatible?
     PaymentType public immutable paymentType;
 
+    // TODO give it a better name
```
kupermind, line 44: @DavidMinarsch Please provide a name, as it will require a significant change across all the contracts.

contracts/OlasMech.sol
DavidMinarsch: we'll have a call about this

contracts/OlasMech.sol
DavidMinarsch: ok

contracts/OlasMech.sol
DavidMinarsch: yes still thinking

PR #62, opened 2024-12-23
refactor and fix: addressing recent changes

kupermind:
- Addressing recent changes.

### DIFF
--- a/contracts/BalanceTrackerFixedPriceNative.sol
+++ b/contracts/BalanceTrackerFixedPriceNative.sol
@@ -35,7 +35,14 @@ contract BalanceTrackerFixedPriceNative is BalanceTrackerFixedPriceBase {
         wrappedNativeToken = _wrappedNativeToken;
     }
 
-    function _checkNativeValue() internal virtual override {}
+    function _getOrRestrictNativeValue() internal virtual override returns (uint256) {
+        // Update balance with native value
+        if (msg.value > 0) {
+            emit Deposit(msg.sender, address(0), msg.value);
+        }
+
+        return msg.value;
+    }
 
     function _getRequiredFunds(address, uint256) internal virtual override returns (uint256) {
         return 0;
--- a/contracts/BalanceTrackerFixedPriceToken.sol
+++ b/contracts/BalanceTrackerFixedPriceToken.sol
@@ -17,6 +17,11 @@ interface IToken {
     /// @param amount Amount to transfer to.
     /// @return True if the function execution is successful.
     function transferFrom(address from, address to, uint256 amount) external returns (bool);
+
+    /// @dev Gets the amount of tokens owned by a specified account.
+    /// @param account Account address.
+    /// @return Amount of tokens owned.
+    function balanceOf(address account) external view returns (uint256);
 }
 
 contract BalanceTrackerFixedPriceNative is BalanceTrackerFixedPriceBase {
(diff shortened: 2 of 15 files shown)


contracts/BalanceTrackerFixedPriceNative.sol
```diff
@@ -35,7 +35,14 @@ contract BalanceTrackerFixedPriceNative is BalanceTrackerFixedPriceBase {
         wrappedNativeToken = _wrappedNativeToken;
     }
 
-    function _checkNativeValue() internal virtual override {}
+    function _getOrRestrictNativeValue() internal virtual override returns (uint256) {
```
kupermind, line 38: This is the current refactor of incoming payment.

contracts/BalanceTrackerFixedPriceToken.sol
```diff
@@ -38,16 +43,29 @@ contract BalanceTrackerFixedPriceNative is BalanceTrackerFixedPriceBase {
...
         if (msg.value > 0) {
             revert NoDepositAllowed(msg.value);
         }
+
+        return 0;
     }
 
     function _getRequiredFunds(address requester, uint256 balanceDiff) internal virtual override returns (uint256) {
-        // TODO check balances before and after
+        uint256 balanceBefore = IToken(olas).balanceOf(address(this));
```
kupermind, line 56: Additional check, although with OLAS it's pretty much safe by default.

contracts/MechFactoryFixedPriceNative.sol
```diff
@@ -37,12 +50,19 @@ contract MechFactoryFixedPriceNative {
         // Decode max delivery rate
         uint256 maxDeliveryRate = abi.decode(payload, (uint256));
 
+        uint256 localNonce = _nonce;
         // Get salt
-        bytes32 salt = keccak256(abi.encode(block.timestamp, msg.sender, serviceId));
+        bytes32 salt = keccak256(abi.encode(block.timestamp, msg.sender, serviceId, localNonce));
```
kupermind, line 55: Adding nonce as per audit suggestion

contracts/MechFixedPriceToken.sol
```diff
@@ -5,13 +5,16 @@ import {OlasMech} from "./OlasMech.sol";
 
 /// @title MechFixedPriceToken - Smart contract for OlasMech that accepts a fixed price payment for services in native token.
 contract MechFixedPriceToken is OlasMech {
+    // keccak256(FixedPriceToken) = 3679d66ef546e66ce9057c4a052f317b135bc8e8c509638f7966edfd4fcf45e9
+    bytes32 public constant PAYMENT_TYPE = 0x3679d66ef546e66ce9057c4a052f317b135bc8e8c509638f7966edfd4fcf45e9;
```
kupermind, line 9: Hashing approach always wins :). Although in this case it feels like we maybe need to call it FixedPriceOlas? Either way is fine.

contracts/MechMarketplace.sol
```diff
@@ -468,24 +467,24 @@ contract MechMarketplace is IErrorsMarketplace {
         // Decrease the number of undelivered requests
         numUndeliveredRequests--;
         // Increase the amount of requester delivered requests
-        mapDeliveryCounts[requester]++;
+        mapDeliveryCounts[mechDelivery.requester]++;
         // Increase the amount of mech delivery counts
-        mapAgentMechDeliveryCounts[msg.sender]++;
+        mapMechDeliveryCounts[msg.sender]++;
         // Increase the amount of mech service multisig delivered requests
         mapMechServiceDeliveryCounts[mechServiceMultisig]++;
```
kupermind, line 474: We still need both - one is for mech statistics, another is for multisig activity tracker.

contracts/MechMarketplace.sol
```diff
@@ -602,41 +593,5 @@ contract MechMarketplace is IErrorsMarketplace {
             }
         }
     }
-
-    /// @dev Gets the requests count for a specific account.
-    /// @param account Account address.
-    /// @return Requests count.
-    function getRequestsCount(address account) external view returns (uint256) {
```
kupermind, line 609: Default ABI is enough

contracts/MechFixedPriceToken.sol
DavidMinarsch: hm I think the token can be different kinds for the same factory contract (different factory instance) no?

DavidMinarsch (approved):

Sfgangloff (approved):

contracts/MechFixedPriceToken.sol
kupermind: Makes sense, maybe encode it on a level of factory? There we can combine `FixedPriceToken` + token address or anything else.

DavidMinarsch: Hm, I'll think about it.

PR #70, opened 2025-01-02
refactor: withdraw of requester balances is scratched

kupermind:
- Withdraw of requester balances is scratched.

### DIFF
--- a/contracts/BalanceTrackerFixedPriceBase.sol
+++ b/contracts/BalanceTrackerFixedPriceBase.sol
@@ -268,27 +268,4 @@ abstract contract BalanceTrackerFixedPriceBase {
     function processPayment() external returns (uint256, uint256) {
         return _processPayment(msg.sender);
     }
-
-    /// @dev Withdraws funds for a specific requester account.
-    function withdraw() external {
-        // Reentrancy guard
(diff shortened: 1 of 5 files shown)


contracts/BalanceTrackerFixedPriceBase.sol
```diff
@@ -268,27 +268,4 @@ abstract contract BalanceTrackerFixedPriceBase {
     function processPayment() external returns (uint256, uint256) {
         return _processPayment(msg.sender);
     }
-
-    /// @dev Withdraws funds for a specific requester account.
-    function withdraw() external {
```
kupermind, line 273: This was affecting only requester balances, so can just be safely deleted.

DavidMinarsch: Where was the request for this - please link for ease of review

kupermind: Actually it says "context deleted", so not sure now if it's still applicable?
https://docs.google.com/document/d/1K7pD_ipEaxdsyTWwnWXrIoPptsH7qXSPV7HmHHdmxPA/edit?disco=AAABa31QMuM

PR #74, opened 2025-01-09
refactor: initial off-chain requests handling

kupermind:
- Off-chain requests handling.

### DIFF
--- a/contracts/BalanceTrackerBase.sol
+++ b/contracts/BalanceTrackerBase.sol
@@ -38,6 +38,8 @@ error UnauthorizedAccount(address account);
 error TransferFailed(address token, address from, address to, uint256 amount);
 
 abstract contract BalanceTrackerBase {
+    event RequesterBalanceAdjusted(address indexed requester, uint256 deliveryRate, uint256 balance);
+    event MechBalanceAdjusted(address indexed mech, uint256 deliveryRate, uint256 balance);
     event MechPaymentCalculated(address indexed mech, uint256 indexed requestId, uint256 deliveryRate, uint256 rateDiff);
     event Deposit(address indexed account, address indexed token, uint256 amount);
     event Withdraw(address indexed account, address indexed token, uint256 amount);
@@ -73,29 +75,29 @@ abstract contract BalanceTrackerBase {
         buyBackBurner = _buyBackBurner;
     }
 
-    /// @dev Adjusts initial requester balance accounting for max request delivery rate (debit).
+    /// @dev Adjusts initial requester balance accounting for delivery rate (debit).
     /// @param balance Initial requester balance.
-    /// @param maxDeliveryRate Max delivery rate.
+    /// @param deliveryRate Delivery rate.
     function _adjustInitialBalance(
         address requester,
         uint256 balance,
-        uint256 maxDeliveryRate,
+        uint256 deliveryRate,
         bytes memory
     ) internal virtual returns (uint256) {
         // Check the request delivery rate for a fixed price
-        if (balance < maxDeliveryRate) {
+        if (balance < deliveryRate) {
             // Get balance difference
-            uint256 balanceDiff = maxDeliveryRate - balance;
+            uint256 balanceDiff = deliveryRate - balance;
             // Adjust balance
             balance += _getRequiredFunds(requester, balanceDiff);
         }
 
-        if (balance < maxDeliveryRate) {
-            revert InsufficientBalance(balance, maxDeliveryRate);
+        if (balance < deliveryRate) {
+            revert InsufficientBalance(balance, deliveryRate);
         }
 
         // Adjust account balance
-        return (balance - maxDeliveryRate);
+        return (balance - deliveryRate);
     }
 
     /// @dev Adjusts final requester balance accounting for possible delivery rate difference (debit).
@@ -173,11 +175,12 @@ abstract contract BalanceTrackerBase {
     /// @dev Checks and records delivery rate.
     /// @param requester Requester address.
     /// @param maxDeliveryRate Request max delivery rate.
+    /// @param paymentData Additional payment-related request data, if applicable.
     function checkAndRecordDeliveryRate(
         address requester,
         uint256 maxDeliveryRate,
-        bytes memory
-    ) external payable {
+        bytes memory paymentData
+    ) public payable {
         // Reentrancy guard
         if (_locked > 1) {
             revert ReentrancyGuard();
@@ -196,8 +199,10 @@ abstract contract BalanceTrackerBase {
         uint256 balance = mapRequesterBalances[requester] + initAmount;
 
         // Adjust account balance
-        balance = _adjustInitialBalance(requester, balance, maxDeliveryRate, "");
+        balance = _adjustInitialBalance(requester, balance, maxDeliveryRate, paymentData);
         mapRequesterBalances[requester] = balance;
+
+        emit RequesterBalanceAdjusted(requester, maxDeliveryRate, balance);
 
         _locked = 1;
     }
@@ -208,6 +213,12 @@ abstract contract BalanceTrackerBase {
     /// @param requestId Request Id.
     /// @param maxDeliveryRate Requested max delivery rate.
     function finalizeDeliveryRate(address mech, address requester, uint256 requestId, uint256 maxDeliveryRate) external {
+        // Reentrancy guard
+        if (_locked > 1) {
+            revert ReentrancyGuard();
+        }
+        _locked = 2;
+
         // Check for marketplace access
         if (msg.sender != mechMarketplace) {
             revert MarketplaceOnly(msg.sender, mechMarketplace);
@@ -223,12 +234,13 @@ abstract contract BalanceTrackerBase {
 
         // Check for delivery rate difference
         uint256 rateDiff;
+        uint256 balance;
         if (maxDeliveryRate > actualDeliveryRate) {
             // Return back requester overpayment debit / credit
             rateDiff = maxDeliveryRate - actualDeliveryRate;
 
             // Adjust requester balance
-            uint256 balance = _adjustFinalBalance(requester, rateDiff);
+            balance = _adjustFinalBalance(requester, rateDiff);
             mapRequesterBalances[requester] = balance;
         } else {
             // Limit the rate by the max chosen one as that is what the requester agreed on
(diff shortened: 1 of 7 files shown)


contracts/MechMarketplace.sol
```diff
@@ -494,6 +511,135 @@ contract MechMarketplace is IErrorsMarketplace {
         _locked = 1;
     }
 
+    /// @dev Verifies provided request hash against its signature.
+    /// @param requester Requester address.
+    /// @param requestHash Request hash.
+    /// @param signature Signature bytes associated with the signed request hash.
+    function _verifySignedHash(address requester, bytes32 requestHash, bytes memory signature) internal view {
```
77ph, line 518: This implementation is not the same as `permit` https://github.com/transmissions11/solmate/blob/e8f96f25d48fe702117ce76c79228ca4f20206cb/src/tokens/ERC20.sol#L116

1. No deadline
2. The main point of the permit is that the hash is calculated on the fly and includes both chainID & uniq nonce
_verifySignedHash(address requester, bytes32 requestHash, bytes memory signature)
So the requester submits raw data and a signature (in off-chain signature calculated based on on-onchain algo including chain-id & uniq nonce). The signature is a one-time thing. You can't use even a valid signature twice.

kupermind: Synced on and modified

contracts/BalanceTrackerBase.sol
```diff
@@ -73,29 +75,29 @@ abstract contract BalanceTrackerBase {
         buyBackBurner = _buyBackBurner;
     }
 
-    /// @dev Adjusts initial requester balance accounting for max request delivery rate (debit).
+    /// @dev Adjusts initial requester balance accounting for delivery rate (debit).
     /// @param balance Initial requester balance.
-    /// @param maxDeliveryRate Max delivery rate.
```
kupermind, line 78: This function is used for any arbitrary delivery rate, not just a single max one.

contracts/BalanceTrackerBase.sol
```diff
@@ -196,9 +199,11 @@ abstract contract BalanceTrackerBase {
         uint256 balance = mapRequesterBalances[requester] + initAmount;
 
         // Adjust account balance
-        balance = _adjustInitialBalance(requester, balance, maxDeliveryRate, "");
```
kupermind, line 199: This must be an actual variable, not a hardcoded empty arg.

contracts/BalanceTrackerBase.sol
```diff
@@ -208,6 +213,12 @@ abstract contract BalanceTrackerBase {
     /// @param requestId Request Id.
     /// @param maxDeliveryRate Requested max delivery rate.
     function finalizeDeliveryRate(address mech, address requester, uint256 requestId, uint256 maxDeliveryRate) external {
+        // Reentrancy guard
+        if (_locked > 1) {
```
kupermind, line 217: Need to guard as overridden internals might eventually call externals.

contracts/BalanceTrackerBase.sol
```diff
@@ -223,22 +234,49 @@ abstract contract BalanceTrackerBase {
...
         }
 
         // Record payment into mech balance
-        mapMechBalances[mech] += actualDeliveryRate;
+        balance = mapMechBalances[mech];
+        balance += actualDeliveryRate;
+        mapMechBalances[mech] = balance;
 
         emit MechPaymentCalculated(mech, requestId, actualDeliveryRate, rateDiff);
+        emit MechBalanceAdjusted(mech, actualDeliveryRate, balance);
```
kupermind, line 256: Also add emit with mech balances change

contracts/BalanceTrackerBase.sol
```diff
@@ -223,22 +234,49 @@ abstract contract BalanceTrackerBase {
...
+    /// @param requester Requester address.
+    /// @param totalDeliveryRate Total batch delivery rate.
+    /// @param paymentData Additional payment-related request data, if applicable.
+    function adjustRequesterMechBalances(
+        address mech,
+        address requester,
+        uint256 totalDeliveryRate,
+        bytes memory paymentData
+    ) external payable {
+        checkAndRecordDeliveryRate(requester, totalDeliveryRate, paymentData);
```
kupermind, line 272: This seems to fit nicely to update with pre-payment, if required.

contracts/MechMarketplace.sol
```diff
@@ -334,14 +354,14 @@ contract MechMarketplace is IErrorsMarketplace {
 
     /// @dev Registers a request.
     /// @notice The request is going to be registered for a specified priority mech.
-    /// @param data Self-descriptive opaque data-blob.
+    /// @param requestData Self-descriptive opaque request data-blob.
```
kupermind, line 357: Was confusing before

contracts/MechMarketplace.sol
```diff
@@ -414,21 +434,18 @@ contract MechMarketplace is IErrorsMarketplace {
...
 
     /// @dev Delivers a request.
     /// @notice This function can only be called by the mech delivering the request.
     /// @param requestId Request id.
-    /// @param requestData Self-descriptive opaque data-blob.
-    function deliverMarketplace(
-        uint256 requestId,
-        bytes memory requestData
-    ) external {
+    /// @param deliveryData Self-descriptive opaque delivery data-blob.
```
kupermind, line 447: Was confusing before

contracts/MechMarketplace.sol
```diff
@@ -489,7 +506,154 @@ contract MechMarketplace is IErrorsMarketplace {
...
+        emit MarketplaceDeliver(priorityMech, msg.sender, mechDelivery.requester, requestId, deliveryData);
+
+        _locked = 1;
+    }
+
+    /// @dev Verifies provided request hash against its signature.
+    /// @param requester Requester address.
+    /// @param requestHash Request hash.
+    /// @param signature Signature bytes associated with the signed request hash.
+    function _verifySignedHash(address requester, bytes32 requestHash, bytes memory signature) internal view {
```
kupermind, line 518: Mostly taken from our autonolas-registries, but modified the order, such that the majority of contracts follow the standard of calling `isValidSignature()` function.

contracts/MechMarketplace.sol
```diff
@@ -489,7 +506,154 @@ contract MechMarketplace is IErrorsMarketplace {
...
+        // Increase mech karma that delivers the request
+        IKarma(karma).changeMechKarma(msg.sender, int256(requestDatas.length));
+
+        // Get balance tracker address
+        bytes32 mechPaymentType = IMech(msg.sender).paymentType();
+        address balanceTracker = mapPaymentTypeBalanceTrackers[mechPaymentType];
+
+        // Process payment
+        IBalanceTracker(balanceTracker).adjustRequesterMechBalances{value: msg.value}(msg.sender, requester,
+            totalDeliveryRate, paymentData);
```
kupermind, line 651: I believe we should still give a chance to pre-pay via a mech if there's not enough credit?

contracts/MechMarketplace.sol
```diff
@@ -489,7 +506,161 @@ contract MechMarketplace is IErrorsMarketplace {
...
+            revert ReentrancyGuard();
+        }
+        _locked = 2;
+        
+        // Check mech
+        address mechServiceMultisig = checkMech(msg.sender);
+
+        // TODO array length checks
+
+        // TODO check requester?
```
kupermind, line 612: Can requester be a service multisig as well?

contracts/BalanceTrackerBase.sol
```diff
@@ -223,22 +234,49 @@ abstract contract BalanceTrackerBase {
...
+        balance += actualDeliveryRate;
+        mapMechBalances[mech] = balance;
 
         emit MechPaymentCalculated(mech, requestId, actualDeliveryRate, rateDiff);
+        emit MechBalanceAdjusted(mech, actualDeliveryRate, balance);
+
+        _locked = 1;
+    }
+
+    /// @dev Adjusts requester and mech balances for direct batch request processing.
```
DavidMinarsch, line 261: who can call this? can we add @notice?

contracts/MechMarketplace.sol
```diff
@@ -43,8 +56,11 @@ contract MechMarketplace is IErrorsMarketplace {
     event SetMechFactoryStatuses(address[] mechFactories, bool[] statuses);
     event SetPaymentTypeBalanceTrackers(bytes32[] paymentTypes, address[] balanceTrackers);
     event MarketplaceRequest(address indexed requester, address indexed requestedMech, uint256 requestId, bytes data);
-    event MarketplaceDeliver(address indexed priorityMech, address indexed actualMech, address indexed requester,
+    event MarketplaceDeliver(address indexed priorityMech, address indexed deliveryMech, address indexed requester,
         uint256 requestId, bytes data);
+    event MarketplaceDeliverBatch(address indexed priorityMech, address indexed deliveryMech, address indexed requester,
```
DavidMinarsch, line 61: Not related to this event: how is a race condition prevented between delivering mechs that have all compete for the same on-chain balance of the requester?

Similarly, how does priority vs delivering mech distinction work, if at all?

contracts/MechMarketplace.sol
```diff
@@ -489,7 +506,161 @@ contract MechMarketplace is IErrorsMarketplace {
...
+
+        // Final check is for the requester address itself
+        if (recRequester != requester) {
+            revert WrongRequesterAddress(recRequester, requester);
+        }
+    }
+
+    /// @dev Approves request hash for requester address.
+    /// @param hash Provided request hash to approve.
+    function requesterApproveHash(bytes32 hash) external {
```
DavidMinarsch, line 580: what is the usecase for this?

contracts/MechMarketplace.sol
DavidMinarsch: Ideally yes, we should see how this is handled in Safe

contracts/MechMarketplace.sol
DavidMinarsch: I don't understand what the usecase would be. Why would the mech pre-pay? They already had the off-chain cost of doing this, why would they now pay themselves?

contracts/MechMarketplace.sol
kupermind: If called via Safe, for example, such that the path of not going via `isValidSignature()` is executed. We had a long debate with Andrey about it, and ultimately decided to leave this scenario as it's supported by Safe.

contracts/MechMarketplace.sol
kupermind: Ah yes, I keep forgetting to check for requests not being executed on-chain.

kupermind: Here the priority mech is strictly a delivery mech, as it's already delivering requests. The idea is:
- Mech collects signed requests;
- Checks that the user has enough balance to cover delivery (easy to do off-chain by checking user balance);
- Processes requests;
- Provides raw input request data and signatures as proofs;
- Provides corresponding delivery requests data and delivery rates.

To be on a safe side it still makes sense to remove requester ability to withdraw their funds.

kupermind:
@DavidMinarsch re your last comment:

>  I don't understand what the usecase would be. Why would the mech pre-pay? They already had the off-chain cost of doing this, why would they now pay themselves?

If requester didn't have enough balance on Mech Marketplace, is there a possibility to relay value via the mech, such that the requester balance increases and becomes sufficient to pay for the mech work? Or we consider all balances final before executing requests and the requester either has enough balances to pay for or doesn't.

contracts/MechMarketplace.sol
kupermind: If requester didn't have enough balance on Mech Marketplace, is there a possibility to relay value via the mech, such that the requester balance increases and becomes sufficient to pay for the mech work? Or we consider all balances final before executing requests and the requester either has enough balances to pay for or doesn't.

contracts/MechMarketplace.sol
kupermind: Had another long discussion about this, let's remove this ambiguous scenario. Correct Safe supports `isValidSignature()` to follow the standard, so we need to consider the general case of a contract following the EIP-1271 workflow for signature verification.

contracts/MechMarketplace.sol
kupermind: ok just talked to Ardian, this is super silly :) I will modify it as obviously mech acts without a requester to ask for additional funds.

contracts/MechMarketplace.sol
kupermind: Talked to Ardian, he also favors ban on withdraws for requesters.

PR #75, opened 2025-01-10
chore: deployment scripts

kupermind:
- Deployment scripts.


contracts/mechs/nevermined/BalanceTrackerNvmSubscription.sol
```diff
@@ -34,35 +39,29 @@ error Overflow(uint256 provided, uint256 max);
...
-        if (_subscriptionNFT == address(0)) {
-            revert ZeroAddress();
-        }
-
-        if (_subscriptionTokenId == 0) {
-            revert ZeroValue();
-        }
-
-        subscriptionNFT = _subscriptionNFT;
-        subscriptionTokenId = _subscriptionTokenId;
```
kupermind, line 65: Subscription is set separately as it needs to be tied to this contract address

contracts/mechs/nevermined/BalanceTrackerNvmSubscription.sol
```diff
@@ -34,35 +39,29 @@ error Overflow(uint256 provided, uint256 max);
...
 
     // Subscription NFT
-    address public immutable subscriptionNFT;
+    address public subscriptionNFT;
     // Subscription token Id
-    uint256 public immutable subscriptionTokenId;
+    uint256 public subscriptionTokenId;
+
+    // Temporary owner address
+    address public owner;
```
kupermind, line 55: Temporary ownership is dropped when the subscription is set

contracts/mechs/nevermined/BalanceTrackerNvmSubscription.sol
```diff
@@ -130,6 +129,26 @@ contract BalanceTrackerNvmSubscription is BalanceTrackerBase {
         return 0;
     }
 
+    /// @dev Process mech payment.
+    /// @param mech Mech address.
+    /// @return mechPayment Mech payment.
+    function _processPayment(address mech) internal virtual override returns (uint256 mechPayment, uint256) {
```
kupermind, line 135: As far as the understanding goes, all the fees are already accounted for when the subscription is paid for and mech gets its cleared cut.

DavidMinarsch (approved):

PR #76, opened 2025-01-10
refactor: requests and deliveries in bulk and optimization

kupermind:
- Requests and deliveries in bulk and optimization.


contracts/BalanceTrackerBase.sol
```diff
@@ -39,8 +39,7 @@ error TransferFailed(address token, address from, address to, uint256 amount);
 
 abstract contract BalanceTrackerBase {
     event RequesterBalanceAdjusted(address indexed requester, uint256 deliveryRate, uint256 balance);
-    event MechBalanceAdjusted(address indexed mech, uint256 deliveryRate, uint256 balance);
-    event MechPaymentCalculated(address indexed mech, uint256 indexed requestId, uint256 deliveryRate, uint256 rateDiff);
```
kupermind, line 43: One event is enough, otherwise they duplicate each other

contracts/BalanceTrackerBase.sol
```diff
@@ -55,7 +54,7 @@ abstract contract BalanceTrackerBase {
     // Collected fees
     uint256 public collectedFees;
     // Reentrancy lock
-    uint256 internal _locked = 1;
```
kupermind, line 58: Transient changes state during a transaction only, and then rolls back to the previous value. It also consumes only 100 gas to change the value.

contracts/BalanceTrackerBase.sol
```diff
@@ -1,9 +1,16 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.28;
 
-import {IMech} from "./interfaces/IMech.sol";
+interface IMech {
```
kupermind, line 4: No more interaction with the mech except for mech multisig verification.

contracts/BalanceTrackerBase.sol
```diff
@@ -196,95 +204,129 @@ abstract contract BalanceTrackerBase {
         // Get account balance
         uint256 balance = mapRequesterBalances[requester] + initAmount;
 
+        // Total requester delivery rate is number of requests coming to a selected mech
+        uint256 totalDeliveryRate = deliveryRate * numRequests;
+
```
kupermind, line 209: This is calculated as part of default batch behavior.

contracts/BalanceTrackerBase.sol
```diff
@@ -196,95 +204,129 @@ abstract contract BalanceTrackerBase {
...
+        locked = true;
 
         // Check for marketplace access
         if (msg.sender != mechMarketplace) {
             revert MarketplaceOnly(msg.sender, mechMarketplace);
         }
 
-        // Get actual delivery rate
-        uint256 actualDeliveryRate = IMech(mech).getFinalizedDeliveryRate(requestId);
+        uint256 numRequests = deliveredRequests.length;
```
kupermind, line 241: Process only requests that were delivered. This bool array is created in the mech marketplace to reflect which request Ids were actually delivered by the mech. This allows to process bulks and not get blocked by requests that were already delivered.

contracts/BalanceTrackerBase.sol
```diff
@@ -196,95 +204,129 @@ abstract contract BalanceTrackerBase {
...
-            revert ZeroValue();
+        // Get total mech and requester delivery rates
+        uint256 totalMechDeliveryRate;
+        uint256 totalRequesterDeliveryRate;
+        uint256 totalRateDiff;
+        for (uint256 i = 0; i < numRequests; ++i) {
+            // Check if request was delivered
+            if (deliveredRequests[i]) {
+                totalMechDeliveryRate += mechDeliveryRates[i];
+                totalRequesterDeliveryRate += requesterDeliveryRates[i];
```
kupermind, line 252: Mechs have their final say in their price when delivering a request, whereas requesters have posted requests agreeing to pay specified mech price. Checks for mechs not exceeding those prices are done in the marketplace before this call.

contracts/BalanceTrackerBase.sol
```diff
@@ -196,95 +204,129 @@ abstract contract BalanceTrackerBase {
...
+                // Check for delivery rate difference
+                uint256 rateDiff;
+                if (requesterDeliveryRates[i] > mechDeliveryRates[i]) {
+                    // Return back requester overpayment debit / credit
+                    rateDiff = requesterDeliveryRates[i] - mechDeliveryRates[i];
+                    totalRateDiff += rateDiff;
+
+                    // Adjust requester balance
+                    balance = _adjustFinalBalance(requesters[i], rateDiff);
+                    mapRequesterBalances[requesters[i]] = balance;
```
kupermind, line 263: This is the same code as before

contracts/BalanceTrackerBase.sol
```diff
@@ -196,95 +204,129 @@ abstract contract BalanceTrackerBase {
...
+    ) external virtual {
+        // Reentrancy guard
+        if (locked) {
+            revert ReentrancyGuard();
+        }
+        locked = true;
+
+        // Get total mech delivery rate
+        uint256 totalMechDeliveryRate;
+        for (uint256 i = 0; i < mechDeliveryRates.length; ++i) {
```
kupermind, line 300: This is now simplified for requests with signatures: simply calculate all the mech prices, take from requester and give to a delivery mech.

contracts/BalanceTrackerBase.sol
```diff
@@ -307,56 +347,29 @@ abstract contract BalanceTrackerBase {
...
-        _locked = 2;
+        locked = true;
 
         (mechPayment, marketplaceFee) = _processPayment(msg.sender);
-
-        _locked = 1;
-    }
-
-    /// @dev Withdraws funds for a specific requester account.
-    function withdraw() external {
```
kupermind, line 341: No more withdraws for requesters, this time finalized.

contracts/MechFixedPriceBase.sol
```diff
@@ -18,13 +18,21 @@ abstract contract MechFixedPriceBase is OlasMech {
     /// @dev Performs actions before the delivery of a request.
     /// @param data Self-descriptive opaque data-blob.
     /// @return requestData Data for the request processing.
-    function _preDeliver(address, uint256, bytes memory data) internal virtual override returns (bytes memory requestData) {
```
kupermind, line 21: Requester addresses are out of the equation for mechs themselves - things are much clearer now.

contracts/OlasMech.sol
```diff
@@ -90,118 +87,91 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
-        // Next element of the zero element will be the newly created element
-        requestIds[1] = requestId;
-        // Previous element of the current next element will be the newly created element
-        mapRequestIds[curNextRequestId][0] = requestId;
+        uint256 numRequests = requestIds.length;
+
+        for (uint256 i = 0; i < requestIds.length; ++i) {
+            uint256 requestId = requestIds[i];
+
+            // TODO optimize or remove?
```
kupermind, line 109: Still thinking it's a good idea to register all requests on the side of a mech, as this way mechs will not be able to deliver requests they have not registered by the marketplace (except for requests with signatures).

contracts/OlasMech.sol
```diff
@@ -90,118 +87,91 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
-        numUndeliveredRequests--;
-
-        // Remove delivered request Id from the request Ids map
-        uint256[2] memory requestIds = mapRequestIds[requestId];
-        // Check if the request Id is invalid (non existent or delivered): previous and next request Ids are zero,
-        // and the zero's element previous request Id is not equal to the provided request Id
-        if (requestIds[0] == 0 && requestIds[1] == 0 && mapRequestIds[0][0] != requestId) {
-            revert RequestIdNotFound(requestId);
-        }
+    function _cleanRequestInfo(uint256 requestId) internal virtual {
```
kupermind, line 135: Just dropping requests now, the stats are updated in the external one in bulk.

contracts/OlasMech.sol
```diff
@@ -90,118 +87,91 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
+        // Traverse requests
+        for (uint256 i = 0; i < numRequests; ++i) {
+            uint256 requestId = requestIds[i];
+
+            // Get request Id from the request Ids map
+            uint256[2] memory requestIdLinks = mapRequestIds[requestId];
+
+            // Check if the request Id is invalid (non existent or delivered): previous and next request Ids are zero,
+            // and the zero's element previous request Id is not equal to the provided request Id
+            if (requestIdLinks[0] == 0 && requestIdLinks[1] == 0 && mapRequestIds[0][0] != requestId) {
```
kupermind, line 168: This is key for registering requests with the mech as mentioned above - non-registered one will not relay to the marketplace.

contracts/OlasMech.sol
```diff
@@ -32,14 +32,11 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
     // Number of total deliveries by this mech
     uint256 public numTotalDeliveries;
     // Reentrancy lock
-    uint256 internal _locked = 1;
+    bool transient locked;
 
-    // Map of undelivered requests counts for corresponding addresses in this agent mech
-    mapping(address => uint256) public mapUndeliveredRequestsCounts;
```
kupermind, line 38: I believe this is not needed at all. Undelivered requests for specific requesters can be found in marketplace records / events.

contracts/OlasMech.sol
```diff
@@ -32,14 +32,11 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
+    // TODO Check if needed as requests are checked by Marketplace
     // Cyclical map of request Ids
     mapping(uint256 => uint256[2]) public mapRequestIds;
-    // Map of request Id => sender address
-    mapping(uint256 => address) public mapRequestAddresses;
```
kupermind, line 42: This is also obsolete - mech should not care about requester addresses.

contracts/OlasMech.sol
```diff
@@ -268,27 +217,56 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
+
+        // Preliminary delivery processing
+        bytes[] memory deliveryDatas = _prepareDeliveries(requestIds, datas);
+
+        // Get finalized delivery rates
+        uint256[] memory deliveryRates = getFinalizedDeliveryRates(requestIds);
+
+        // Mech marketplace delivery finalization
+        // Some of deliveries might be front-run by other mechs, and thus only actually delivered ones are recorded
+        bool[] memory deliveredRequests = IMechMarketplace(mechMarketplace).deliverMarketplace(requestIds,
```
kupermind, line 245: This is key - get exactly requests that were delivered via the marketplace, account for them, and ignore all undelivered ones.

contracts/OlasMech.sol
```diff
@@ -338,6 +316,7 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
         return multisig == getOperator();
     }
 
+    // TODO remove?
```
kupermind, line 319: This function seems to be overly complicated, but we can leave it not to perform those calculations via the script, if needed.

contracts/MechMarketplace.sol
```diff
@@ -172,7 +171,7 @@ contract MechMarketplace is IErrorsMarketplace {
         uint256 newMaxResponseTimeout
     ) internal {
         // Check for zero values
-        if (newFee == 0 || newMinResponseTimeout == 0 || newMaxResponseTimeout == 0) {
```
kupermind, line 175: As discussed - marketplace fee can be set to zero.

contracts/OlasMech.sol
```diff
@@ -90,118 +87,91 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
     function _request(
         address requester,
-        uint256 requestId,
-        bytes memory data
+        uint256[] memory requestIds,
+        bytes[] memory datas
     ) internal virtual {
-        mapUndeliveredRequestsCounts[requester]++;
-        // Record the requestId => sender correspondence
-        mapRequestAddresses[requestId] = requester;
```
kupermind, line 114: As specified above, no need to track requesters by mechs, all this info is in the MM.

contracts/OlasMech.sol
```diff
@@ -90,118 +87,91 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
-
-        // Get the mech delivery info from the mech marketplace
-        IMechMarketplace.MechDelivery memory mechDelivery =
-            IMechMarketplace(mechMarketplace).mapRequestIdDeliveries(requestId);
-
-        // Instantly return with empty data if the request has been delivered
-        // This allows not to fail batch requests transactions
-        if (mechDelivery.deliveryMech != address(0)) {
-            return "";
-        }
```
kupermind, line 178: No need to pull the actual request info from the MM and then cancel it. The delivered request statuses for each request Id is going to be pulled when calling MM delivery function. Then, the undelivered requests are cleaned up.

contracts/OlasMech.sol
```diff
@@ -90,118 +87,91 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
             }
-            // Note, revoking the request for the priority mech happens later via revokeRequest
-        } else {
-            // The requester is non-zero if it is delivered by the priority mech
-            _cleanRequestInfo(requester, requestId);
-        }
 
-        // Check for max delivery rate compared to requested one
-        if (maxDeliveryRate > mechDelivery.deliveryRate) {
-            revert Overflow(maxDeliveryRate, mechDelivery.deliveryRate);
```
kupermind, line 195: This is now done on the MM side when trying to deliver requests.

contracts/OlasMech.sol
```diff
@@ -268,27 +217,56 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
 
-        // Request delivery
-        bytes memory requestData = _deliver(requestId, data);
+        // Check array sizes
+        if (requestIds.length == 0 || requestIds.length != datas.length) {
+            revert WrongArrayLength(requestIds.length, datas.length);
+        }
+
+        // Preliminary delivery processing
+        bytes[] memory deliveryDatas = _prepareDeliveries(requestIds, datas);
```
kupermind, line 238: Check for requests existence and pre-deliver them.

contracts/OlasMech.sol
```diff
@@ -268,27 +217,56 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
+        bool[] memory deliveredRequests = IMechMarketplace(mechMarketplace).deliverMarketplace(requestIds,
+            deliveryRates, deliveryDatas);
+
+        uint256 numRequests = requestIds.length;
+        uint256 numDeliveries;
+        // Traverse all requests to select delivered ones
+        for (uint256 i = 0; i < numRequests; ++i) {
+            uint256 requestId = requestIds[i];
+            if (deliveredRequests[i]) {
+                numDeliveries++;
```
kupermind, line 254: If request is delivered, it's summed up, otherwise it's cleaned up.

contracts/MechMarketplace.sol
```diff
@@ -381,132 +380,240 @@ contract MechMarketplace is IErrorsMarketplace {
...
 
-        // Calculate request Id
-        requestId = getRequestId(msg.sender, requestData, mapNonces[msg.sender]);
+        // Allocate set of requestIds
+        requestIds = new uint256[](numRequests);
 
-        // Update requester nonce
-        mapNonces[msg.sender]++;
+        // Get deliveryRate as priority mech max delivery rate
+        uint256 deliveryRate = IMech(priorityMech).maxDeliveryRate();
```
kupermind, line 402: Get mech delivery rate on the MM side at the beginning of the loop, as it's the same price that's used in each request.

contracts/MechMarketplace.sol
```diff
@@ -381,132 +380,240 @@ contract MechMarketplace is IErrorsMarketplace {
...
 
         // Record the request count
-        mapRequestCounts[msg.sender]++;
+        mapRequestCounts[msg.sender] += numRequests;
         // Increase the number of undelivered requests
-        numUndeliveredRequests++;
+        numUndeliveredRequests += numRequests;
         // Increase the total number of requests
-        numTotalRequests++;
+        numTotalRequests += numRequests;
```
kupermind, line 455: Same logic as before, just in bulks

contracts/MechMarketplace.sol
```diff
@@ -381,132 +380,246 @@ contract MechMarketplace is IErrorsMarketplace {
...
+        // Allocate requester related arrays
+        address[] memory requesters = new address[](numRequests);
+        uint256[] memory requesterDeliveryRates = new uint256[](numRequests);
 
-        // Check for request existence
-        if (priorityMech == address(0)) {
-            revert ZeroAddress();
-        }
+        // Traverse all requests being delivered
+        for (uint256 i = 0; i < numRequests; ++i) {
```
kupermind, line 550: Below in the loop it's the same logic as before, but please comment on TODO-s with your feedback. I'm trying to minimize the number of scenarios when mechs have to fail with reverts.

contracts/MechMarketplace.sol
```diff
@@ -381,132 +380,246 @@ contract MechMarketplace is IErrorsMarketplace {
...
+                revert Overflow(mechDeliveryRates[i], mechDelivery.deliveryRate);
+            }
+
+            // If delivery mech is different from the priority one
+            if (priorityMech != msg.sender) {
+                // Within the defined response time only a chosen priority mech is able to deliver
+                if (block.timestamp > mechDelivery.responseTimeout) {
+                    // Decrease priority mech karma as the mech did not deliver
+                    // Needs to stay atomic as each different priorityMech can be any address
+                    IKarma(karma).changeMechKarma(priorityMech, -1);
```
kupermind, line 581: This is the only atomic part left, but there's no other way of doing it.

contracts/MechMarketplace.sol
```diff
@@ -381,132 +380,246 @@ contract MechMarketplace is IErrorsMarketplace {
...
+            mechDelivery.deliveryMech = msg.sender;
+
+            // Record requester
+            requesters[i] = mechDelivery.requester;
+
+            // Increase the amount of requester delivered requests
+            mapDeliveryCounts[requesters[i]]++;
+
+            // Mark request as delivered
+            deliveredRequests[i] = true;
```
kupermind, line 600: This is a key array that's returned to the mech - it allows for it to see which requests were actually delivered and which ones need to be cleaned up as they are not going to be delivered.

contracts/MechMarketplace.sol
```diff
@@ -30,8 +30,8 @@ interface ISignatureValidator {
     function isValidSignature(bytes32 hash, bytes memory signature) external view returns (bytes4 magicValue);
 }
 
-// Mech delivery info struct
-struct MechDelivery {
+// Request info struct
+struct RequestInfo {
```
kupermind, line 34: This naming reflects its purpose much better

contracts/MechMarketplace.sol
```diff
@@ -42,6 +42,8 @@ struct MechDelivery {
     uint256 responseTimeout;
     // Delivery rate
     uint256 deliveryRate;
+    // Payment type
+    bytes32 paymentType;
```
kupermind, line 46: We need to check payment type for priority vs delivery mechs.

contracts/MechMarketplace.sol
```diff
@@ -381,132 +382,258 @@ contract MechMarketplace is IErrorsMarketplace {
...
+        uint256 deliveryRate = IMech(priorityMech).maxDeliveryRate();
+
+        // Get priority mech payment type
+        bytes32 paymentType = IMech(priorityMech).paymentType();
+
+        // Get nonce
+        uint256 nonce = mapNonces[msg.sender];
+
+        // Traverse all requests
+        for (uint256 i = 0; i < requestDatas.length; ++i) {
```
kupermind, line 413: Logic in the loop is same as before, just in bulk and added payment type check

contracts/MechMarketplace.sol
```diff
@@ -381,132 +382,258 @@ contract MechMarketplace is IErrorsMarketplace {
...
+            }
 
-        // Check for request existence
-        if (priorityMech == address(0)) {
-            revert ZeroAddress();
-        }
+            // Check payment type
+            if (requestInfo.paymentType != paymentType) {
+                revert WrongPaymentType(paymentType);
+            }
```
kupermind, line 575: Payment type is checked, and if incorrect - definitely need to revert.

contracts/MechMarketplace.sol
```diff
@@ -381,132 +382,258 @@ contract MechMarketplace is IErrorsMarketplace {
...
         // Increase the amount of mech delivery counts
-        mapMechDeliveryCounts[msg.sender]++;
+        mapMechDeliveryCounts[msg.sender] += numDeliveries;
         // Increase the amount of mech service multisig delivered requests
-        mapMechServiceDeliveryCounts[mechServiceMultisig]++;
+        mapMechServiceDeliveryCounts[mechServiceMultisig] += numDeliveries;
 
         // Increase mech karma that delivers the request
-        IKarma(karma).changeMechKarma(msg.sender, 1);
+        IKarma(karma).changeMechKarma(msg.sender, int256(numDeliveries));
```
kupermind, line 627: Same as before, just in bulks

contracts/MechMarketplace.sol
```diff
@@ -570,20 +697,20 @@ contract MechMarketplace is IErrorsMarketplace {
     /// @param deliveryRates Corresponding set of actual charged delivery rates for each request.
     /// @param deliveryDatas Corresponding set of self-descriptive opaque delivery data-blobs.
     /// @param paymentData Additional payment-related request data, if applicable.
-    function deliverMarketplaceBatchWithSignatures(
+    function deliverMarketplaceWithSignatures(
```
kupermind, line 700: This one is not modified, except for naming. Refer to the deliver with signatures PR.

contracts/MechMarketplace.sol
```diff
@@ -643,22 +769,21 @@ contract MechMarketplace is IErrorsMarketplace {
...
+        IKarma(karma).changeRequesterMechKarma(requester, msg.sender, int256(numRequests));
         // Increase mech karma that delivers the request
-        IKarma(karma).changeMechKarma(msg.sender, int256(requestDatas.length));
+        IKarma(karma).changeMechKarma(msg.sender, int256(numRequests));
 
         // Get balance tracker address
         address balanceTracker = mapPaymentTypeBalanceTrackers[IMech(msg.sender).paymentType()];
 
-        // Process payment
-        IBalanceTracker(balanceTracker).adjustRequesterMechBalances{value: msg.value}(msg.sender, requester,
```
kupermind, line 653: This was wrong - no additional payment can be transferred. Also, total delivery rate calculation happens in the balance tracker side.

contracts/OlasMech.sol
```diff
@@ -2,17 +2,27 @@
 pragma solidity ^0.8.28;
 
 import {IErrorsMech} from "./interfaces/IErrorsMech.sol";
-import {IMechMarketplace} from "./interfaces/IMechMarketplace.sol";
 import {ImmutableStorage} from "../lib/gnosis-mech/contracts/base/ImmutableStorage.sol";
 import {IServiceRegistry} from "./interfaces/IServiceRegistry.sol";
 import {Mech} from "../lib/gnosis-mech/contracts/base/Mech.sol";
 
+/// @dev Mech Marketplace interface
+interface IMechMarketplace {
```
kupermind, line 10: Interface is stripped such that there's only one external function needed.

contracts/OlasMech.sol
```diff
@@ -90,118 +97,88 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
+        uint256 numSelfRequests;
+        // Traverse requests
+        for (uint256 i = 0; i < numRequests; ++i) {
+            uint256 requestId = requestIds[i];
+            // Perform a pre-delivery of the data if it needs additional parsing
+            deliveryDatas[i] = _preDeliver(requestId, datas[i]);
+
+            // Clean request info
+            // Get request Id from the request Ids map
+            uint256[2] memory requestIdLinks = mapRequestIds[requestId];
```
kupermind, line 162: Check if request was assigned to this mech, and if it was - reduce the undelivered count of this mech.

contracts/OlasMech.sol
```diff
@@ -90,118 +97,88 @@ abstract contract OlasMech is Mech, IErrorsMech, ImmutableStorage {
...
         // Increase the total number of requests
-        numTotalRequests++;
-
-        emit Request(requester, requestId, data);
-    }
-
-    /// @dev Cleans the request info from all the relevant storage.
-    /// @param requester Requester address.
-    /// @param requestId Request Id.
-    function _cleanRequestInfo(address requester, uint256 requestId) internal virtual {
```
kupermind, line 141: Became obsolete in this form, as there are no push from MM anymore. This logic went into pre-deliver of requests to count the number of self and not self ones.

Sfgangloff:
I think that in the function _adjustInitialBalance of BalanceTrackerBase, the conditions are wrong. We have twice "if (balance < deliveryRate)", one of the conditions should be "if (balance >= deliveryRate)"

contracts/MechMarketplace.sol
```diff
@@ -381,132 +376,268 @@ contract MechMarketplace is IErrorsMarketplace {
...
+        bytes32 paymentType = IMech(msg.sender).paymentType();
+
+        // Traverse all requests being delivered
+        for (uint256 i = 0; i < numRequests; ++i) {
+            // Get request info struct
+            RequestInfo storage requestInfo = mapRequestIdInfos[requestIds[i]];
+            address priorityMech = requestInfo.priorityMech;
+
+            // TODO continue instead of revert? This would let us just ignore incorrect requests without reverting the bulk
+            // TODO Additional thought: leave revert as it would stop from mech trying to request arbitrary requests and not actually recorded ones
```
DavidMinarsch, line 566: What scenario is this?

contracts/MechMarketplace.sol
```diff
@@ -381,132 +376,268 @@ contract MechMarketplace is IErrorsMarketplace {
...
-            if (block.timestamp > mechDelivery.responseTimeout) {
-                // Decrease priority mech karma as the mech did not deliver
-                IKarma(karma).changeMechKarma(priorityMech, -1);
-                // Revoke request from the priority mech
-                IMech(priorityMech).revokeRequest(requestId);
-            } else {
-                // Priority mech responseTimeout is still >= block.timestamp
-                revert PriorityMechResponseTimeout(mechDelivery.responseTimeout, block.timestamp);
+            // TODO continue instead of revert? This would let us just ignore incorrect requests without reverting the bulk
+            // TODO Additional thought: The price between posting requests and delivering could have changed, thus just continue such that mech cleans those request info on their side
```
DavidMinarsch, line 583: Hm. need to understsnd the scenarios

contracts/MechMarketplace.sol
```diff
@@ -381,132 +376,268 @@ contract MechMarketplace is IErrorsMarketplace {
...
+            // If delivery mech is different from the priority one
+            if (priorityMech != msg.sender) {
+                // Within the defined response time only a chosen priority mech is able to deliver
+                if (block.timestamp > requestInfo.responseTimeout) {
+                    // Decrease priority mech karma as the mech did not deliver
+                    // Needs to stay atomic as each different priorityMech can be any address
+                    IKarma(karma).changeMechKarma(priorityMech, -1);
+                } else {
+                    // TODO continue instead of revert? This would let us just ignore incorrect requests without reverting the bulk
+                    // TODO Additional thought: revert as mech has just miscalculated time, does not mean it has to clear that request
```
DavidMinarsch, line 599: scenarios needed

contracts/MechMarketplace.sol
Sfgangloff: Just one idea. In general when a request in a batch would be reverted as a single request, always continue, add bool where the request is recorded (if it is) which says the request is invalid (invalid requests can't be delivered). Maybe under some conditions on invalid requests found in the batch (like, if they are all found invalid or enough of them) then revert the batch. Then we have a cleaner function which removes data of invalid requests.

contracts/OlasMech.sol
DavidMinarsch: can they not look up the same on the mm?

contracts/OlasMech.sol
DavidMinarsch: yes we should offer simple getters for info off-chain mech needs

contracts/BalanceTrackerBase.sol
kupermind: It comes form the requester to a single mech, so for now the assumption is that those are identical. But things can be overwritten in the future via virtual override.

contracts/OlasMech.sol
kupermind: They could, one could go through events though and filter by mechs, where in the current version each mech holds their full record of undelivered requests.

contracts/MechMarketplace.sol
kupermind: Mech is trying to deliver request that was never registered on the MM

contracts/BalanceTrackerBase.sol
```diff
@@ -196,95 +204,135 @@ abstract contract BalanceTrackerBase {
...
+        emit RequesterBalanceAdjusted(requester, totalDeliveryRate, balance);
 
-        _locked = 1;
+        _locked = false;
     }
 
     /// @dev Finalizes mech delivery rate based on requested and actual ones.
     /// @param mech Delivery mech address.
-    /// @param requester Requester address.
-    /// @param requestId Request Id.
```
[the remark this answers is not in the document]
kupermind: Request Ids are not needed in balance trackers, as no mech calls are needed anymore.

contracts/MechMarketplace.sol
kupermind: Thanks, for the simplicity I'd immediately fail on critical ones and continue on all other ones. I'll discuss with Ardian the flexibility there, but we don't want to complicate the code much more.

contracts/MechMarketplace.sol
kupermind: When user posted a request - the rate was 5. Before request is processed and delivered, the price was increased by the mech operator. Now mech wants a higher rate for a request, but the user has agreed on a smaller rate.

contracts/MechMarketplace.sol
kupermind: Delivery mech is trying to deliver a request during the time when priority mech still has its priority time to deliver.

contracts/BalanceTrackerBase.sol
kupermind: Yes, there are Deposit() events for subscription-like deposit and for deposits on the go. There are events for adjusting mech balances and separately for adjusting requester balances.

PR #80, opened 2025-01-14
chore and test: requestId uint to bytes32, adding signature testing

kupermind:
- `requestId` uint256 to bytes32, adding signature testing.


77ph (approved):

contracts/MechMarketplace.sol
```diff
@@ -803,8 +803,8 @@ contract MechMarketplace is IErrorsMarketplace {
         address account,
         bytes memory data,
         uint256 nonce
-    ) public view returns (uint256 requestId) {
-        requestId = uint256(keccak256(
```
kupermind, line 807: `bytes32` is basic for signatures, so conceptually better to switch to that one. Got green light from launch team.

PR #86, opened 2025-01-20
refactor: no need to check requester service Id

kupermind:
- No need to check requester service Id;
- Delivery rate is added as a hashing parameter as part of `requestId()`;
- Renaming for more generic usage;
- NO CODE CHANGE IN OVERALL LOGIC, JUST ABI UPDATES WHERE NEEDED.


contracts/BalanceTrackerBase.sol
```diff
@@ -59,8 +59,8 @@ abstract contract BalanceTrackerBase {
 
     // Mech marketplace address
     address public immutable mechMarketplace;
-    // Buy back burner address
-    address public immutable buyBackBurner;
```
kupermind, line 63: Drainer is a better name - it's generic.

contracts/MechMarketplace.sol
```diff
@@ -204,14 +204,12 @@ contract MechMarketplace is IErrorsMarketplace {
     /// @notice The request is going to be registered for a specified priority mech.
     /// @param requestDatas Set of self-descriptive opaque request data-blobs.
     /// @param priorityMechServiceId Priority mech service Id.
-    /// @param requesterServiceId Requester service Id, or zero if EOA.
```
kupermind, line 207: This came out as a parameter that can always be equal to zero, and it passes all the checks. So checks on requesters become obsolete.

contracts/MechMarketplace.sol
```diff
@@ -237,9 +235,6 @@ contract MechMarketplace is IErrorsMarketplace {
             revert ZeroAddress();
         }
 
-        // Check requester
-        checkRequester(msg.sender, requesterServiceId);
```
kupermind, line 241: `checkRequester()` ultimately becomes obsolete as well.

contracts/MechMarketplace.sol
```diff
@@ -260,7 +255,7 @@ contract MechMarketplace is IErrorsMarketplace {
             }
 
             // Calculate request Id
-            requestIds[i] = getRequestId(msg.sender, requestDatas[i], nonce);
+            requestIds[i] = getRequestId(msg.sender, requestDatas[i], deliveryRate, nonce);
```
kupermind, line 258: Adding deliveryRate into the requestId is required as this is the part of the request the requester signs on, and it becomes the only verifiable part of hash when requesting off-chain

77ph (approved):

DavidMinarsch (approved):

PR #98, opened 2025-02-06
fix: addressing issue 8

kupermind:
- Addressing issue 8.

### DIFF
--- a/contracts/MechMarketplace.sol
+++ b/contracts/MechMarketplace.sol
@@ -53,6 +53,7 @@ struct RequestInfo {
 /// @author Silvere Gangloff - <silvere.gangloff@valory.xyz>
 contract MechMarketplace is IErrorsMarketplace {
     event CreateMech(address indexed mech, uint256 indexed serviceId, address indexed mechFactory);
+    event RemoveMech(address indexed mech, uint256 indexed serviceId);
     event OwnerUpdated(address indexed owner);
     event ImplementationUpdated(address indexed implementation);
     event MarketplaceParamsUpdated(uint256 fee, uint256 minResponseTimeout, uint256 maxResponseTimeout);
@@ -131,8 +132,8 @@ contract MechMarketplace is IErrorsMarketplace {
     mapping(bytes32 => address) public mapPaymentTypeBalanceTrackers;
     // Mapping of account nonces
     mapping(address => uint256) public mapNonces;
-    // Mapping of service ids to mechs
-    mapping(uint256 => address) public mapServiceIdMech;
+    // Mapping of mech to service id
+    mapping(address => uint256) public mapMechServiceIds;
 
 
     /// @dev MechMarketplace constructor.
@@ -289,7 +290,7 @@ contract MechMarketplace is IErrorsMarketplace {
     /// @param requestDatas Set of self-descriptive opaque request data-blobs.
     /// @param maxDeliveryRate Max delivery rate requester agrees to supply.
     /// @param paymentType Payment type.
-    /// @param priorityMechServiceId Priority mech service Id.
+    /// @param priorityMech Priority mech address.
     /// @param responseTimeout Relative response time in sec.
     /// @param paymentData Additional payment-related request data (optional).
     /// @return requestIds Set of request Ids.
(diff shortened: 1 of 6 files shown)


77ph (approved):

contracts/MechMarketplace.sol
```diff
@@ -913,7 +933,7 @@ contract MechMarketplace is IErrorsMarketplace {
         uint256 mechServiceId = IMech(mech).tokenId();
 
         // Check mech validity as it must be created and recorded via this marketplace
-        if (mapServiceIdMech[mechServiceId] != mech) {
+        if (mapMechServiceIds[mech] != mechServiceId) {
```
Saw-mon-and-Natalie, line 936: a malicious fake `mech` (that is not even registered in the system) can return `0` for `IMech(mech).tokenId()` (`mechServiceId`) and thus bypass the `checkMech(mech)` check.

Saw-mon-and-Natalie: currently one can even simplify the current implementation (since request now provide the mech and not the service id) and get rid of `mapMechServiceIds` and instead just check for `mapAgentMechFactories[mech]` and make sure it is not `address(0)` and in the `remove(mech)` you clear this mapping instead.

kupermind: Good point, fixing

77ph (approved):

contracts/MechMarketplace.sol
```diff
@@ -570,13 +574,31 @@ contract MechMarketplace is IErrorsMarketplace {
...
-        mapServiceIdMech[serviceId] = mech;
+        // Add mapping for mech => service Id
+        mapMechServiceIds[mech] = serviceId;
         numMechs++;
 
         emit CreateMech(mech, serviceId, mechFactory);
     }
 
+    /// @dev Removes mech from current Mech Marketplace.
+    /// @notice Only mech service Id multisig can call this function.
```
mariapiamo, line 585: why the multisig and not the service owner?

contracts/MechMarketplace.sol
```diff
@@ -131,8 +132,8 @@ contract MechMarketplace is IErrorsMarketplace {
     mapping(bytes32 => address) public mapPaymentTypeBalanceTrackers;
     // Mapping of account nonces
     mapping(address => uint256) public mapNonces;
-    // Mapping of service ids to mechs
-    mapping(uint256 => address) public mapServiceIdMech;
+    // Mapping of mech to service id
+    mapping(address => uint256) public mapMechServiceIds;
```
kupermind, line 136: @DavidMinarsch @mariapiamo this mapping does not make sense anymore IMO, and so is the `remove()` function below.

Each mech service Id is accessible directly via `mech->tokenId()`, and the existence of a mech is validated by the whitelisted factory that created it.

contracts/MechMarketplace.sol
```diff
@@ -570,13 +574,31 @@ contract MechMarketplace is IErrorsMarketplace {
...
+        mapMechServiceIds[mech] = serviceId;
         numMechs++;
 
         emit CreateMech(mech, serviceId, mechFactory);
     }
 
+    /// @dev Removes mech from current Mech Marketplace.
+    /// @notice Only mech service Id multisig can call this function.
+    /// @param mech Mech address.
+    function remove(address mech) external {
```
kupermind, line 587: Not needed IMO

mariapiamo (approved):

contracts/MechMarketplace.sol
```diff
@@ -555,6 +553,12 @@ contract MechMarketplace is IErrorsMarketplace {
     /// @param mechFactory Mech factory address.
     /// @return mech The created mech instance address.
     function create(uint256 serviceId, address mechFactory, bytes memory payload) external returns (address mech) {
+        // Check for msg.sender to be a serviceId multisig
+        (, address multisig, , , , , ) = IServiceRegistry(serviceRegistry).mapServices(serviceId);
```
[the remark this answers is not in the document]
DavidMinarsch: For simplicity I'd say not at this point.

contracts/MechMarketplace.sol
DavidMinarsch: OK, then let's not have it

contracts/MechMarketplace.sol
DavidMinarsch: OK re removing the mapping.

However, we should still require service id being passed for "create". Otherwise we decouple marketplace from registry, which is not something we wanted to do at this stage. We want to ensure each mech is backed by a service registered on our registry.

Issue #121, opened 2025-07-30
Correct indexed arrays in events


kupermind:
At least one indexed array in contract events has been observed:
https://github.com/valory-xyz/ai-registry-mech/blob/766fe81dc2320dc5fa35a4e2e285f9142b4c72ca/contracts/MechMarketplace.sol#L64

This leads to events emitting hash instead of array values. Reference: https://www.tribagus.com/understanding-indexed-string-in-solidity-events

Need to correct this and other possible events having the same issue.

PR #122, opened 2025-08-04
refactor: events modification in MM

kupermind:
- Events modification in Mech Marketplace.
- Batch request is going to have array of request data-s emitted.
- Batch delivery with signatures is going to have request data along with delivery data. Note that it's only possible to read single events for each delivery.
- Batch onchain delivery: nothing is possible to change without a subsequent modification of Mech contract.
- Addressing issue #121.

### DIFF
--- a/contracts/MechMarketplace.sol
+++ b/contracts/MechMarketplace.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.28;
+pragma solidity ^0.8.30;
 
 import {IErrorsMarketplace} from "./interfaces/IErrorsMarketplace.sol";
 import {IBalanceTracker} from "./interfaces/IBalanceTracker.sol";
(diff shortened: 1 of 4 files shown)


contracts/MechMarketplace.sol
```diff
@@ -60,11 +60,11 @@ contract MechMarketplace is IErrorsMarketplace {
...
     event SetPaymentTypeBalanceTrackers(bytes32[] paymentTypes, address[] balanceTrackers);
     event MarketplaceRequest(address indexed priorityMech, address indexed requester, uint256 numRequests,
-        bytes32[] requestIds);
-    event MarketplaceDelivery(address indexed deliveryMech, address[] indexed requesters, uint256 numDeliveries,
+        bytes32[] requestIds, bytes[] requestDatas);
+    event MarketplaceDelivery(address indexed deliveryMech, address[] requesters, uint256 numDeliveries,
         bytes32[] requestIds, bool[] deliveredRequests);
     event Deliver(address indexed mech, address indexed mechServiceMultisig, bytes32 requestId, uint256 deliveryRate,
-        bytes data);
+        bytes requestData, bytes deliveryData);
```
kupermind, line 67: `Deliver` event for delivery with signatures: adding request and delivery data-s

contracts/MechMarketplace.sol
```diff
@@ -206,7 +206,7 @@ contract MechMarketplace is IErrorsMarketplace {
     function _deliverMarketplaceWithSignatures(
         address requester,
         bytes32 paymentType,
-        DeliverWithSignature[] calldata deliverWithSignatures,
+        DeliverWithSignature[] memory deliverWithSignatures,
```
kupermind, line 209: This change is needed for addressing the "stack too deep" error otherwise. I don't see an issue here as the function is internal.

contracts/MechMarketplace.sol
```diff
@@ -299,7 +299,7 @@ contract MechMarketplace is IErrorsMarketplace {
         bytes32 paymentType,
         address priorityMech,
         uint256 responseTimeout,
-        bytes calldata paymentData
+        bytes memory paymentData
```
kupermind, line 302: This change is needed for addressing the "stack too deep" error otherwise. I don't see an issue here as the function is internal.

contracts/MechMarketplace.sol
```diff
@@ -60,11 +60,11 @@ contract MechMarketplace is IErrorsMarketplace {
     event SetMechFactoryStatuses(address[] mechFactories, bool[] statuses);
     event SetPaymentTypeBalanceTrackers(bytes32[] paymentTypes, address[] balanceTrackers);
     event MarketplaceRequest(address indexed priorityMech, address indexed requester, uint256 numRequests,
-        bytes32[] requestIds);
-    event MarketplaceDelivery(address indexed deliveryMech, address[] indexed requesters, uint256 numDeliveries,
+        bytes32[] requestIds, bytes[] requestDatas);
```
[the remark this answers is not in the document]
77ph: lgtm
@kupermind @mariapiamo 
As an idea to avoid problems in the future, let's add reading of important events to the tests.

PR #148, opened 2026-03-16
doc: x402 spec, off-chain marketplace API spec, analytics ETL spec + schema

kupermind:
## Summary

Docs-only PR. Lays out two parallel workstreams and the design needed to ship either of them. **No contracts in this PR.** All on-chain and service-side code lands in follow-ups.

The PR started as learnings on x402 and has expanded into the full set of design documents that line up with the implementation plan:

1. **x402 payment family** — a new payment family alongside the existing three (Native, Token, NVM Subscription). Adds HTTP-native paid API access via EIP-3009 `transferWithAuthorization` for USDC. Decision-guide pair includes an MPP session alternative for the same problem.
2. **Off-chain marketplace migration** — moves mech requests off the public chain rails while keeping them billable and auditable. Two phases plus an optional third. Companion analytics ETL service replaces three separate metrics pipelines (olas-website, townhall-kpis, mech-predict) with one pre-computed source served from a dedicated Postgres behind a Wildcard FastAPI.

### x402 spec set

| File | Purpose |
|------|---------|
| `docs/x402_spec.md` | Full x402 compatibility spec. Defines `BalanceTrackerX402USDC`, the EIP-3009 settlement override, the dual-signature pattern, and protocol conformance. |
| `docs/x402_implementation_plan.md` | Phased build plan, spec review, and remaining open items for x402. |
| `docs/x402scan_integration.md` | How a mech becomes discoverable on x402scan / Bazaar. Self-registration via `register-origin` + `/.well-known/x402`. |
| `docs/x402_pre_audit.md` | Pre-implementation review of x402 against the existing contracts. |
| `docs/x402_key_findings_pre_audit.md` | Severity-ranked summary of pre-audit findings. |
| `docs/mpp_session_spec.md` | Alternative: MPP session spec. Defines `MppEscrow`, `BalanceTrackerMppSession`, voucher verification, and channel lifecycle. |
| `docs/x402_vs_mpp.md` | Decision guide between the two payment families. |

### Off-chain marketplace migration spec set

| File | Purpose |
|------|---------|
| `docs/marketplace_api_spec.md` | The marketplace-side API and rollout spec. Phase 1 builds (off-chain HTTP path, commit-reveal privacy, structured 402 + Payment-Receipt, mech-side write path into the data lake, analytics ETL service, metrics Postgres + public Wildcard API, olas-website rewrites, historical IPFS ETL). Phase 2 rolls out without leaving an analytics gap. Phase 3 is deferred scaling. |
| `docs/mech_analytics_etl_spec.md` | Phase 1 sub-phase 2: the Python ETL service, the pipeline jobs (score new rows, late resolution sweep, roll-ups, FPMM trades ingest), the metric math ported from `mech-predict/benchmark/scorer.py`, and the FastAPI architecture. |
| `docs/mech_analytics_etl_schema.md` | Table-schema reference for the metrics Postgres. Six tables (`per_request_scores`, `tool_aggregates`, `agent_aggregates` with rolling + daily snapshots, `mech_aggregates`, `chain_aggregates`, `cursor_state`), full DDL, indexes, and which Wildcard API endpoint each table backs. |

### Project guide

| File | Purpose |
|------|---------|
| `CLAUDE.md` | Project guide for code assistants working in this repo, with pointers to the new workstream docs. |

## What's been addressed since the first review pass

- Contracts renamed from `BalanceTrackerX402` / `MechFixedPriceTokenX402` / `MechFactoryFixedPriceTokenX402` to `BalanceTrackerX402USDC` / `MechFixedPriceTokenX402USDC` / `MechFactoryFixedPriceTokenX402USDC` so the v1 ERC-20 (EIP-3009 USDC) is explicit in the name, matching the existing `MechFixedPriceTokenUSDC` convention. The PAYMENT_TYPE hash was already `keccak256("X402USDC")`.
- `/supported` and `/health` facilitator endpoints dropped from the collapsed ResourceServer+Facilitator=Mech architecture. The 402 body's `accepts[]` already advertises supported config; the mech's existing healthcheck owns the 503-on-settlement-timeout signal. Only `POST /verify` remains.
- Batching framing corrected: the spec now describes that batching IS achievable at the SC level (`paymentData` decodes as an `Authorization[]`, `_adjustInitialBalance` loops `transferWithAuthorization`). MechMarketplace and OlasMech stay untouched. v1 still recommends per-request settlement for failure isolation; batching is a later optimization.
- Rename of the rollout doc from `mech_prepay_spec.md` to `marketplace_api_spec.md` so the filename describes what the document scopes.
- Analytics ETL spec added, plus a dedicated table-schema reference that reflects the latest scope decisions (daily snapshots on `agent_aggregates`, three Wildcard API endpoints under `/v1/metrics/...`, 14-day late-resolution window).

## What the spec pair looks like

```
              Existing                                    New (pick ONE)
┌──────────────────────────────────┐          ┌─────────────────────────────────┐
│ Native · Token · NVM Subscription │   +     │ x402   OR   MPP session         │
└──────────────────────────────────┘          └─────────────────────────────────┘
                              ↓
                  MechMarketplace.mapPaymentTypeBalanceTrackers
                              ↓
                  (existing pattern, no marketplace changes)
```

```
Off-chain marketplace migration

mech ─HTTP─► request/response stays off the public chain rails
mech ─signed batch─► predict-api data lake (mech_requests, mech_responses)
                          │
                          ▼
                    Analytics ETL ─reads─► Omen + Polymarket subgraphs
                          │
                          ▼
                    Metrics Postgres (per_request_scores + 5 aggregates)
                          │
                          ▼
                    Wildcard FastAPI under /v1/metrics/
                          │
       ┌──────────────────┼────────────────────────┐
       ▼                  ▼                        ▼
  olas-website      townhall-kpis           mech-predict
```

## Out of scope for this PR

- All Solidity for both payment options (lives in a follow-up).
- Mech-side off-chain implementation for x402 and for the request/response HTTP path (lives in the mech repo).
- The ETL service implementation itself (new `mech-analytics-etl` repo).
- Consumer rewrites (olas-website, townhall-kpis, mech-predict daily report).

## Test plan

- [ ] Docs reviewed against the contracts (`MechMarketplace.sol`, `OlasMech.sol`, `BalanceTrackerBase.sol`, `BalanceTrackerFixedPriceToken.sol`) for accuracy of `paymentData` flow, `_adjustInitialBalance` extension hook, and the marketplace fee model.
- [ ] Pre-audit findings cross-checked against the spec; resolved findings marked struck-through in `x402_key_findings_pre_audit.md`.
- [ ] Decision guide reviewed for consistency with both spec docs.
- [ ] ETL schema doc validated against the latest scope (windows, daily snapshots, endpoint surface).


docs/x402_spec.md
```diff
@@ -0,0 +1,489 @@
...
+- Registered via: MechMarketplace owner sets `mapPaymentTypeBalanceTrackers[keccak256("X402USDC")] = BalanceTrackerX402 address`.
+- **MechMarketplace: ZERO changes. OlasMech: ZERO changes. Existing BalanceTrackers: ZERO changes.**
+
+#### Required contract set
+
+The marketplace resolves the balance tracker via `mapPaymentTypeBalanceTrackers[mech.paymentType()]`. For x402 to route correctly, **three** contracts are required (following the existing `MechFixedPriceTokenUSDC` pattern):
+
+| Contract | Purpose |
+|----------|---------|
+| `BalanceTrackerX402` | EIP-3009 payment settlement (overrides `_adjustInitialBalance`) |
```
DavidMinarsch: Maybe we use:

`BalanceTrackerX402USDC`
`MechFixedPriceTokenX402USDC`
`MechFactoryFixedPriceTokenX402USDC`
or similar, to also disambiguite the erc20/3009 used still?

DavidMinarsch: Although, not sure why "FixedPrice" and not just `MechTokenX402USDC`?

docs/x402_spec.md
```diff
@@ -0,0 +1,489 @@
...
+
+### 3.4 Mech as Facilitator
+
+Rather than deploying a separate facilitator service, the mech itself exposes the facilitator endpoints. This keeps the architecture simple: the mech handles EIP-3009 signature verification, nonce checks, and balance checks in-process before executing the tool. There is no external service dependency and no cross-service communication for the verification step.
+
+**Facilitator endpoints exposed by the mech:**
+
+- `POST /verify` — validates the EIP-3009 signature, client balance, nonce, and timestamps. Returns `{ valid: true/false, reason }`.
+- `GET /supported` — returns supported payment configurations (see Section 3.5 for schema).
+- `GET /health` — liveness check. Returns 503 after repeated settlement timeouts (see Section 3.6).
```
DavidMinarsch: Who is meant to be calling these in our ResourceServer+Facilitator=Mech architecture?

docs/x402_spec.md
```diff
@@ -0,0 +1,489 @@
...
+
+**Quote Generation**
+- For fixed-price mechs (including x402): `quote = maxDeliveryRate + (maxDeliveryRate * fee_bps / 10000)`. No tool execution needed.
+- For NVM/dynamic mechs (future scope): run tool with `delivery_rate=0` to get cost estimate, then inflate with fee.
+- Query `MechMarketplace.fee()` for current fee bps.
+
+**In-Process Verification (Facilitator)**
+- `POST /verify` — decode `X-Payment` header, validate EIP-3009 signature, check client on-chain USDC balance >= quote, check nonce not in local set, check `validBefore` is in future. Return `{ valid, reason }`.
+- `GET /supported` — return `{ kinds: [{ x402Version: 1, scheme: "exact", network: "<chain>" }] }` per chain the mech supports
+- `GET /health` — liveness check; returns 503 after 3 consecutive settlement timeouts (see Section 3.6)
```
DavidMinarsch: see above-why needed in this collapsed architecture?

docs/x402_spec.md
DIvyaNautiyal07: I guess supported and health is an overkill since we already have healthcheck,  but we do need verify which validates the signature

docs/x402_implementation_plan.md
```diff
@@ -0,0 +1,254 @@
...
+```solidity
+// Loop only sums rates
+for (uint256 i = 0; i < mechDeliveryRates.length; ++i) {
+    totalMechDeliveryRate += mechDeliveryRates[i];
+}
+// Called ONCE with total, not per-request
+requesterBalance = _adjustInitialBalance(requester, requesterBalance, totalMechDeliveryRate, paymentData);
+```
+
+This means a single `paymentData` (single EIP-3009 authorization) must cover the **total** of all delivery rates in the batch. The spec's recommendation for per-request settlement (single-element arrays) is correct and is the only practical approach for x402 — but the stated reason ("called once per request") is wrong.
```
DavidMinarsch: Does this mean we cannot batch?

kupermind: The short answer is: batching works, but each batch carries a single paymentData (one EIP-3009 authorization) that must cover the summed total of all delivery rates in the
  batch.

  Looking at BalanceTrackerBase.sol:306-320, the loop sums mechDeliveryRates[] into totalMechDeliveryRate, then calls _adjustInitialBalance once with that total — not once per
  request. So a batch of 5 requests at 10200 each would need one EIP-3009 authorization for 51000.

  This means batching is technically possible if:
  1. The client signs a single EIP-3009 authorization covering the total amount upfront
  2. The mech aggregates multiple requests from the same requester into one deliverMarketplaceWithSignatures call

  However, this is impractical for x402 because:
  - Each x402 request arrives independently via HTTP with its own X-Payment header and its own EIP-3009 signature
  - The client doesn't know at signing time how many requests will be batched together, so they can't sign for the aggregate amount
  - You'd need a way to combine multiple independent EIP-3009 authorizations into a single paymentData — but the contract only calls _adjustInitialBalance once, so only one
  transferWithAuthorization executes

  The practical approach (already recommended in the spec) is per-request settlement: single-element mechDeliveryRates[] arrays, each with its own paymentData. This also isolates
   failures — since deliverMarketplaceWithSignatures reverts the entire transaction on any failure, batching risks one bad payment killing unrelated deliveries.

  The spec text in Section 3.3 should be corrected: it says _adjustInitialBalance "is called once per request inside the batch loop," which is wrong — it's called once per
  adjustMechRequesterBalances invocation with the total. The recommendation is right, the stated reason is wrong.
  
  Correcting ^

docs/x402_implementation_plan.md
```diff
@@ -0,0 +1,239 @@
...
+
+#### 2.1 Create contract test file
+
+**Location:** `test/MechFixedPriceTokenX402.js`
+
+**Test cases:**
+1. **Happy path:** Client signs EIP-3009, mech delivers via `deliverMarketplaceWithSignatures` with `paymentData`, `_adjustInitialBalance` override executes `transferWithAuthorization`, funds move atomically, balances update correctly
+2. **Fallback:** Empty `paymentData` falls back to standard `transferFrom` pre-deposit flow via `super._adjustInitialBalance`
+3. **Validation failures:** Wrong `from` address, wrong `to` address, insufficient `value`, expired `validBefore`, already-used nonce
+4. **Per-request settlement:** Single-element arrays with individual EIP-3009 authorizations
```
mariapiamo: with the  EIP-3009 authorization, does the client assume that at a certain point the mech will respond?
What happens if the mech directly use the payment authorization without actually delivery nothing off-chain and on-chain?

docs/x402_spec.md
```diff
@@ -0,0 +1,489 @@
...
+quote = maxDeliveryRate + (maxDeliveryRate * fee_bps / 10000)
+```
+
+The mech queries `MechMarketplace.fee()` for fee bps and returns this quote in the HTTP 402 response body.
+
+**NVM/dynamic-price mechs (future scope):** on requests without payment headers, run tool with `delivery_rate=0` to get cost estimate, then inflate: `quote = tool_cost + (tool_cost * fee_bps / 10000)`.
+
+#### Batch semantics
+
+In `adjustMechRequesterBalances` (BalanceTrackerBase.sol), the batch loop sums `mechDeliveryRates[]` into `totalMechDeliveryRate`, then `_adjustInitialBalance` is called **once** with that total — not once per request. This means a single `paymentData` (single EIP-3009 authorization) must cover the summed total of all delivery rates in the batch. Batching is therefore impractical for x402: each HTTP request arrives independently with its own `X-Payment` header and EIP-3009 signature, and the client cannot know the aggregate amount at signing time. **Per-request settlement (single-element arrays) is the only practical approach.**
```
mariapiamo: but so far mechs heavily uses Batch. 
- Does this mean that mech codebase should be updated a lot? @DIvyaNautiyal07 
- How we handle parallelized tasks if we cannot batch? @DIvyaNautiyal07

docs/x402_implementation_plan.md
DIvyaNautiyal07: The EIP-3009 authorization's to field is the BalanceTrackerX402 contract, not the mech's wallet. The mech can't redirect funds to itself.If the mech tries to execute transferWithAuthorization directly (outside of deliverMarketplaceWithSignatures), USDC arrives at the BalanceTracker but mapRequesterBalances isn't updated — the funds are stuck in the contract with no accounting. The mech can't withdraw them.

The only way for the mech to get paid is through the proper flow: call deliverMarketplaceWithSignatures with valid requestId + data (actual delivery) → marketplace processes delivery → calls adjustMechRequesterBalances → credits mapMechBalances[mech]. The delivery is recorded on-chain and verifiable.So the mech could grief a client (burn their authorization into the contract uselessly), but it gains nothing from doing so. And doing it would tank its Karma score, making it unpickable for future requests.

The validBefore timestamp also limits the authorization window — if the mech doesn't settle within that window, the authorization expires and can't be used.

docs/x402_spec.md
DIvyaNautiyal07: Yes, batching goes away for x402 requests. Each request arrives with its own EIP-3009 authorization that has an expiry window (validBefore), so we can't hold them to accumulate a batch. This means: each x402 request is settled individually via deliverMarketplaceWithSignatures with single-element arrays.

Impact on the mech codebase:

- The batch cycle changes from "collect N requests, settle once" to "settle each x402 request individually
- Gas cost per request increases since we lose batch amortization
- Parallelized tool execution isn't affected, tools still run in parallel, only the settlement changes
- Existing pre-deposit requests still batch normally, only x402 requests lose batching

docs/x402_implementation_plan.md
mariapiamo: @DIvyaNautiyal07  I am not saying that funds can go directly to the mech wallet, I am reporting the following scenario:

- client send a request
- mech sends quotes
- client accepts sending authorization
-  mech 'prentend' to claim funds by sending invalid or null response both off-chain and  on-chain
- mech balance is adjusted, then mech can claim funds
- user gets only invalid/null response

mariapiamo: so, we need to exiplicitely state that: mech are assumed to be trustworthy and there is no mechanism protecting clients for output quality

DIvyaNautiyal07: yes I agree, but this is how it works even now, this is not specific to x402, even now we somehow have to trust mechs and there is no mechanism to protect against it.

docs/x402_implementation_plan.md
DavidMinarsch: That makes it very expensive to settle x402 then.

DavidMinarsch: what would need to change on SC level to support batching?

docs/x402scan_integration.md
```diff
@@ -0,0 +1,295 @@
...
+
+## How does a service end up on x402scan?
+
+There are **two paths**:
+
+### Path 1 — Direct registration
+
+x402scan has a self-registration page: [x402scan.com/resources/register](https://www.x402scan.com/resources/register)
+
+1. Submit the mech endpoint URL (e.g. `https://mech.autonolas.network/predict`)
```
DavidMinarsch: What endpoint is this? it doesn't exist.

docs/x402scan_integration.md
```diff
@@ -0,0 +1,295 @@
...
+2. x402scan hits it
+3. If it returns a valid HTTP 402 response with proper x402 schema, it's listed automatically
+4. No manual approval, no external facilitator required
+
+### Path 2 — Bazaar (automatic, via external facilitator)
```
DavidMinarsch, line 28: We don't want the external faciliator route as it does not allow for marketplace usage. so remove this one

docs/x402scan_integration.md
```diff
@@ -0,0 +1,295 @@
...
+
+The mech is its own facilitator everywhere (as the spec currently describes), and we register on x402scan manually. This is the simplest approach and works on all chains including Gnosis.
+
+### How it works
+
+```
+Same on every chain:
+  Mech verifies EIP-3009 signatures in-process (inside the behaviour)
+  Mech settles via deliverMarketplaceWithSignatures
+  Mech registered on x402scan via register page
```
DavidMinarsch: https://www.x402scan.com/discovery please add to this spec how we'd enable this auto-discovery

docs/x402scan_integration.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+No external facilitator dependency. No per-chain config. No cost.
+
+### Getting on x402scan
```
DavidMinarsch: ideally this is done programmatically by the mech at launch once

docs/x402scan_integration.md
```diff
@@ -0,0 +1,295 @@
...
+
+1. Submit the URL (e.g. `https://mech.autonolas.network/predict`)
+2. x402scan hits it, gets a valid HTTP 402 response
+3. Service is listed — no approval needed
+
+This works for **any chain** including Gnosis, because x402scan validates the response schema, not the chain.
+
+### Limitations
+
+- **No Bazaar auto-indexing** — the mech appears on x402scan but NOT in the Bazaar `/discovery/resources` API. AI agents using the Bazaar to discover services programmatically won't find it.
```
DavidMinarsch: can one be listed on bazaar without using coinbase facilitator?

docs/x402scan_integration.md
```diff
@@ -0,0 +1,295 @@
...
+- **Manual registration per endpoint** — each mech URL needs to be registered individually. New mechs don't auto-appear.
```
DavidMinarsch: see above; should be possible to do programmatically

docs/x402scan_integration.md
```diff
@@ -0,0 +1,295 @@
...
+- **Less ecosystem integration** — x402 clients that use facilitators for discovery won't see the mech. Listed on x402scan's website but not programmatically discoverable.
```
DavidMinarsch: why not programmatically discoverable?

docs/x402scan_integration.md
DIvyaNautiyal07: This endpoint doesn't exist yet, it's a placeholder for the HTTP endpoint the mech will host for receiving x402 payment requests.

docs/x402scan_integration.md
DIvyaNautiyal07: We'd only use the external facilitator for /verify (signature validation), never /settle. Settlement still goes through our deliverMarketplaceWithSignatures → _adjustInitialBalance flow, the marketplace is fully in the loop. The facilitator is just a verification oracle, not a settlement path. But happy to remove if we want to keep things simpler.

docs/x402scan_integration.md
DIvyaNautiyal07: added here - [0b07629](https://github.com/valory-xyz/autonolas-marketplace/pull/148/commits/0b07629a673deae109a2bcafbc94cb592fd8067f)

docs/x402scan_integration.md
DIvyaNautiyal07: Agreed, the mech can call the registration API programmatically at startup:

POST https://x402scan.com/api/x402/registry/register-origin
Body: { "origin": "https://<mech-host>" }
Combined with serving /.well-known/x402 (static JSON listing payable routes), this makes the mech self-registering. Will add this to the startup behaviour.

docs/x402scan_integration.md
DIvyaNautiyal07: The Bazaar spec is open. any facilitator can implement /discovery/resources. However, x402scan hardcodes which facilitators it crawls (currently 27, including Coinbase and PayAI). Getting our own facilitator added would require a PR to the [x402scan repo](https://github.com/Merit-Systems/x402scan).
The simpler path is x402scan's own auto-discovery: the mech serves /.well-known/x402 and calls register-origin at startup. This doesn't go through the Bazaar but achieves the same result, the mech is indexed and discoverable on x402scan.

docs/x402scan_integration.md
DIvyaNautiyal07: Updated the mech now self-registers at startup via POST register-origin. No manual step needed.

docs/x402scan_integration.md
DIvyaNautiyal07: It is programmatically discoverable, x402scan's register-origin + /.well-known/x402 makes the mech queryable via x402scan's API. The limitation was specifically about the Bazaar /discovery/resources endpoint (which is tied to whichever facilitators x402scan crawls currently 27 hardcoded ones). But for practical purposes, x402scan's own discovery is sufficient. Removed this section as it was misleading.

docs/x402_implementation_plan.md
kupermind: **Nothing in MechMarketplace or OlasMech. Only the `BalanceTrackerX402` override needs to handle it.**

The key insight is that `paymentData` is an opaque `bytes` blob — only the specific BalanceTracker interprets it. The existing flow already works for batching:

1. `deliverMarketplaceWithSignatures` accepts `bytes calldata paymentData` (single blob)
2. `adjustMechRequesterBalances` sums `mechDeliveryRates[]` into `totalMechDeliveryRate`
3. `_adjustInitialBalance` is called **once** with that total + the opaque `paymentData`

So `BalanceTrackerX402._adjustInitialBalance` just needs to decode `paymentData` as an **array** of EIP-3009 authorizations instead of a single one:

```solidity
function _adjustInitialBalance(
    address requester,
    uint256 balance,
    uint256 deliveryRate,       // already the summed total
    bytes memory paymentData
) internal override returns (uint256) {
    // Decode array of EIP-3009 authorizations
    Authorization[] memory auths = abi.decode(paymentData, (Authorization[]));
    
    uint256 totalTransferred;
    for (uint256 i = 0; i < auths.length; i++) {
        // Validate each: from == requester, to == address(this)
        IUSDC(token).transferWithAuthorization(
            auths[i].from, auths[i].to, auths[i].value,
            auths[i].validAfter, auths[i].validBefore,
            auths[i].nonce, auths[i].v, auths[i].r, auths[i].s
        );
        totalTransferred += auths[i].value;
    }
    
    if (balance + totalTransferred < deliveryRate) {
        revert InsufficientBalance(balance + totalTransferred, deliveryRate);
    }
    
    return (balance + totalTransferred - deliveryRate);
}
```

**Off-chain side:** The mech accumulates multiple x402 requests (each with its own EIP-3009 auth from the HTTP `X-Payment` header), then packs all the auths into a single `paymentData = abi.encode(authArray)` and calls `deliverMarketplaceWithSignatures` as a normal batch.

**Cost savings:** Each `transferWithAuthorization` is still ~50k gas, but the marketplace overhead (signature verification, karma updates, request bookkeeping) is amortized across the batch — one tx instead of N.

**Bottom line:** Zero interface changes. The opaque `bytes paymentData` design already anticipated this.

docs/x402scan_integration.md
DavidMinarsch: It creates another dependency and the resource server can verify themselves, so not sure why the integration would make sense

docs/x402scan_integration.md
DavidMinarsch: Please then come up with a sensible url schema

kupermind (commented):
## Review

Direction: approve. Two new payment families behind distinct `paymentType` keys, both via the `_adjustInitialBalance` override, with `MechMarketplace`/`OlasMech`/fee logic untouched — this is exactly the seam the contracts were designed for. Decision guide (`x402_vs_mpp.md`) is well-structured and honest about trade-offs.

A few must-fix and clarify items before any Solidity lands. Findings below are grouped by severity.

### Code-checked claims (these all hold)

- `_adjustInitialBalance` is the correct override point. Declared `internal virtual` at `BalanceTrackerBase.sol:90`, called once per batch from both `checkAndRecordDeliveryRates` (`:215`) and `adjustMechRequesterBalances` (`:320`), with the **summed** `totalDeliveryRate` / `totalMechDeliveryRate`. Single-call-with-sum is correct (Issue A resolution holds).
- Fee formula at `BalanceTrackerBase.sol:158` is ceil-rounded over accumulated `mapMechBalances[mech]`, not per-request: `marketplaceFee = (balance * fee + (MAX_FEE_FACTOR - 1)) / MAX_FEE_FACTOR`. L-1 dismissal is correct. (Note: `MAX_FEE_FACTOR = 10_000` is also at `MechMarketplace.sol:88`; current live `fee = 0` on every chain.)
- `getRequestId` mech binding at `MechMarketplace.sol:232` is `getRequestId(msg.sender, requester, requestData, deliveryRate, paymentType, nonce)` — first arg is the calling mech. N-3 fix is correct.
- Subgraph claim holds: `MarketplaceDeliveryWithSignatures` (`MechMarketplace.sol:68`/`:284`) emits both `deliveryMech` and `requester`, and `Deliver` (`:66`/`:258`) carries the mech-side detail. No indexer changes needed.
- `paymentData` flows through unchanged from `MechMarketplace` → `BalanceTracker` (`MechMarketplace.sol:864`).

### M-2 BLOCKER — Gnosis bridged USDC EIP-3009 — RESOLVED in this review

`docs/x402_key_findings_pre_audit.md` flags this as the biggest blocker; one `cast` away. Did it now against `0x2a22f9c3b484c3629090FeED35F17Ff8F88f76F0` on Gnosis:

- `version() = "2"` (matches Circle USDC v2 ABI)
- `DOMAIN_SEPARATOR()` returns a valid value (EIP-712 present)
- `authorizationState(address,bytes32) returns(bool)` — callable, returns `false` for an unused nonce (this is the EIP-3009 nonce-tracking entrypoint)
- `transferWithAuthorization(...)` selector `0xe3ee160e` is dispatched (bare-selector call reverts with `0x` — typical of a function reached without valid args, not "function not found")

**Conclusion**: Gnosis bridged USDC supports EIP-3009. Please downgrade M-2 from BLOCKER → NOTE and update `x402_implementation_plan.md` Open Item 2 / `x402_pre_audit.md` M-3 accordingly. Gnosis is a viable v1 x402 target.

### Must-fix before contracts

1. **MPP `closeChannel` access control — public is wrong as default** (`mpp_session_spec.md` §3.2). Public close lets anyone with a leaked latest voucher prematurely finalise the channel before the mech has settled, leaving the mech to write off accumulated cumulative it can no longer push through `settle()` (because `finalized = true`). The escrow's sig check stops *forged* closes, not *premature* ones. Default should be **onlyOperator-of-an-active-mech-for-this-paymentType**, or at minimum **onlyMech** (any registered mech). Public could be a v2 opt-in. Please pick and pin in v1; this is one of the "undecided in v1" items the spec lists at the end of §8.

2. **MPP channel ⇄ mech binding must be enforced, not "operational hygiene"** (`mpp_session_spec.md` §8 — "Channel identity"). The escrow's `channelId = keccak256(payer, BalanceTracker, token, salt, signer, address(this), chainId)` does not mention the mech. The spec proposes `salt = keccak256(abi.encode(mechAddress))` as a **client SDK convention**. That isn't enough: a buggy or hostile client can open one channel and route payments through a second mech, and the second mech has no way to detect the channel was intended for the first mech. The `BalanceTrackerMppSession` (or the mech behaviour) must reject vouchers for `channelId != expected(mech)`. Either enforce on-chain by including `mech` in `channelId`, or enforce off-chain by having the mech reconstruct the expected `channelId` from `salt = keccak256(msg.sender)` and revert mismatch. Document the enforcement, don't leave it to the client.

3. **Decide quote policy (A vs B) before any contracts** (`x402_spec.md` §3.3, `mpp_session_spec.md` §3.6, `x402_vs_mpp.md` §9.3). Both specs recommend Policy A (mech absorbs fee), both worked examples are computed in Policy B (10200 at 2% fee). Pick one, rewrite both worked examples consistently. With the proposal_01 fee bump to **15%** (see PR #161), the difference between Policy A and Policy B is no longer cosmetic — Policy A means a mech keeps `0.85 * maxDeliveryRate` per request; Policy B means clients are quoted `~1.176 * maxDeliveryRate`. Client-SDK and quote-display work depends on this.

### Should-fix

4. **Batch x402 — operational hazard underplayed** (`x402_spec.md` §3.3 "Batch semantics" + "Critical: atomic batch revert"). The combination of (i) per-request EIP-3009 single-use, (ii) atomic full-revert on any failed delivery, and (iii) tools already executed before settle, means batched x402 carries a real risk of losing N tool-executions to one bad authorization. The recommendation "per-request settlement for v1" is right; please make the *reason* (already-executed tools, no partial recovery) explicit, not just "simpler off-chain".

5. **Defensive `block.chainid` on `BalanceTrackerX402` constructor** (H-2 dismissal in `x402_key_findings_pre_audit.md`). The dismissal relies on Circle deploying USDC at distinct addresses per chain. For bridged USDC (Gnosis), that's not a Circle-controlled invariant. Storing `block.chainid` as immutable in the constructor and asserting in `_adjustInitialBalance` is one line of code and gas-free at deploy time; the H-2 removal is too absolute given the existence of bridged variants. Keep it as belt-and-braces.

6. **Settlement race — recommend a `maxDeliveryRate` ceiling for x402** (`x402_key_findings_pre_audit.md` H-1). Spec accepts the loss bounded by `maxDeliveryRate`. With 15% fee in mind, a $1 `maxDeliveryRate` is `$1` of loss per failed settle — acceptable; `$100/req` becomes meaningful very fast. Recommend a documented x402 ceiling (e.g. `maxDeliveryRate <= $1`) and steer higher-unit-price mechs to MPP. This grounds the decision matrix.

7. **MPP `forceClose` UX gap is bigger than §8 admits**. Client needs native gas on the target chain to open *and* to force-close. The decision guide pitches MPP as "lock USDC, sign vouchers, walk away" — but the recovery path requires the client to have ETH/xDAI and to wait 24h. For one-shot agents this is a real regression vs x402. Either acknowledge MPP is not suitable for pure-USDC clients without sponsored gas (recommended), or commit to ERC-4337 sponsorship as in-scope (significant additional work).

8. **`authorizedSigner` field — pick one for v1** (`mpp_session_spec.md` §8 / §3.1). Either fix `authorizedSigner = payer` (drop the field from the EIP-712 surface and tests) or support delegation (add tests #9 in §10.1). Storing the field but not exercising the delegation path is dead weight + an untested code path.

### Nice-to-have

9. **`MppEscrow` immutability statement.** `channelId` includes `address(this)`. Worth a one-liner in `mpp_session_spec.md` §6.1 that the escrow is non-upgradeable by design — any redeploy invalidates every channelId, so don't.
10. **15% fee × MPP big-batch settle.** Worth a note: at 15% fee, a single MPP settle of 10000 USDC takes 1500 USDC to treasury. Not a bug — but mech operators will see one big fee event instead of many small ones; document the cadence.
11. **`x402_pre_audit.md` and `x402_key_findings_pre_audit.md` overlap heavily.** Consider merging into one file post-revisions; the divergent line numbers and `~~strikethroughs~~` are already confusing.
12. **CLAUDE.md** is a net positive. Minor: it claims "`Coverage / npx hardhat coverage`" — fine, but CI runs `./node_modules/.bin/hardhat coverage` (works the same, just worth aligning the command).

### Non-blocking observations

- Strong agreement with the "ship x402 first, add MPP only when a real high-volume client appears" recommendation in `x402_vs_mpp.md` §7. Catches the existing ecosystem, smaller audit surface, lets MPP scope its own concrete v1 client (Optimus or specific Valory partner) rather than building speculatively.
- `MarketplaceDeliveryWithSignatures` events already attribute per-client correctly via `requester` — both specs lean on this and it's accurate.
- The "Mech as Facilitator" pattern is a clean simplification; no external service dependency, in-process verify is fine for v1.

### Suggested next step

Resolve must-fix items 1–3 (close-channel ACL, channel-mech enforcement, quote policy) and incorporate the Gnosis EIP-3009 verification result, then this is ready to land. Items 4–8 can ride into the contract PR(s) so long as they're carried through the implementation plan.

docs/x402_vs_mpp.md
```diff
@@ -0,0 +1,306 @@
...
+- Two audit scopes
+
+The benefit of supporting both:
+- Every persona above can use the mech without compromise
+
+---
+
+## 7. Recommendation
+
+For the Mech Marketplace v1, the practical sequence is:
```
DavidMinarsch, line 253: **The PR no longer tells one coherent v1 story.** This decision guide says "ship x402 first," but `mech_prepay_spec.md` (added later, committed as "the minimum-viable v1 path") recommends shipping mech-prepay with *zero* contract changes and deferring both x402 and MPP. The PR body says "pick x402 or MPP." Three docs, three different v1 answers.

This guide is supposed to be the tiebreaker, yet it never references `mech_prepay_spec.md` and omits it from References §10. Please either fold mech-prepay into this guide as a third option (it's effectively the §7 "Skip both" bullet, now fully specced) or add a single reconciling section that names the v1 path and relates all three.

docs/x402_spec.md
```diff
@@ -0,0 +1,506 @@
...
+| `delivery_rate` in body | Existing off-chain flow (unchanged) |
+| On-chain event | Existing on-chain flow (unchanged) |
+
+---
+
+## 5. Worked Example
+
+A single request with the following setup:
+
+> **Tool cost:** 0.01 USDC | **Marketplace fee:** 2% (200 bps) | **Quote:** 0.0102 USDC
```
DavidMinarsch: **Stale fee assumption — needs a rebase + number sweep.** This worked example (and every fee figure across all the docs) assumes 2% / 200 bps. `main` has since merged proposal_01 activating a **15% fee** (PR #161). At 15% the quote-policy choice stops being cosmetic: Policy A means the mech keeps 0.85× the listed rate; Policy B means the client is quoted ~1.176× it. Recommend rebasing onto current `main` and re-deriving every worked example at the live fee.

docs/x402_spec.md
```diff
@@ -0,0 +1,506 @@
...
+
+**NVM/dynamic-price mechs (future scope):** on requests without payment headers, run tool with `delivery_rate=0` to get cost estimate, then apply the same chosen policy.
+
+#### Batch semantics
+
+In `adjustMechRequesterBalances` (BalanceTrackerBase.sol), the batch loop sums `mechDeliveryRates[]` into `totalMechDeliveryRate`, then `_adjustInitialBalance` is called **once** with that total — not once per request.
+
+**Batching IS achievable.** Because `paymentData` is an opaque `bytes` blob interpreted only by the specific BalanceTracker, `BalanceTrackerX402._adjustInitialBalance` can decode it as an **array** of EIP-3009 authorizations. The override loops through each authorization, calls `transferWithAuthorization` for each, sums the transferred amounts, and checks the total covers `totalMechDeliveryRate`. No changes to `MechMarketplace`, `OlasMech`, or the `paymentData` interface are required.
```
DavidMinarsch: **x402 batching is single-requester only — worth stating explicitly.** `deliverMarketplaceWithSignatures` takes a single `requester` for the whole call (`MechMarketplace.sol:833-834`). So array-encoded `paymentData` can batch multiple requests from **one** client, but cannot settle multiple distinct clients in one tx. As written this section (and the `x402_vs_mpp.md` comparison row) reads as if batching is general. One sentence pinning the single-requester constraint avoids a wrong inference.

docs/x402_spec.md
```diff
@@ -0,0 +1,506 @@
...
+- **Policy B, client pays grossed-up (mech receives listed rate net).** `quote = ceil(maxDeliveryRate * 10000 / (10000 - fee_bps))`. Client pays slightly more than the listed rate, the mech receives `maxDeliveryRate` after fee. Predictable revenue for mech operators, slightly more complex for client SDKs to compute.
+
+Both work. The marketplace contract does not enforce either. **This needs an explicit policy decision before implementation**, and the same decision must be applied consistently to MPP session quotes (see `docs/mpp_session_spec.md`). v1 recommendation: Policy A. It matches the existing flow and avoids the gross-up math in client SDKs.
+
+The mech queries `MechMarketplace.fee()` for fee bps and returns the resulting quote in the HTTP 402 response body. The worked example in Section 5 below uses Policy B for illustration (the 10200 figure is the grossed-up quote at 2% fee); rewrite to Policy A by setting `quote = maxDeliveryRate = 10000` if that policy is chosen.
```
DavidMinarsch: **The quote-grossing formula is internally inconsistent — not just "pick A vs B."** §3.3 Policy B here is `ceil(maxDeliveryRate * 10000 / (10000 - fee_bps))` → **10205** at 2% on 10000 (rate *net* of fee). But §4 step 2, the §4 routing table, and the §5 worked example all use `maxDeliveryRate + maxDeliveryRate*fee_bps/10000` → **10200** (fee *on top*). Different numbers, different semantics. `mpp_session_spec.md` §3.2/§3.6 copies the additive 10200 version too. Pick one formula and make every example agree — at 15% the discrepancy is no longer rounding noise.

DavidMinarsch (commented):
## Review — docs-only, technically accurate, but not yet coherent as a set

Independently verified the load-bearing contract claims against current `main` (the override seam `_adjustInitialBalance`, the `deliverMarketplaceWithSignatures` call shape and its routing through `adjustMechRequesterBalances`, the `getRequestId(mech, …)` mech-binding, the ceil-rounded fee over accumulated `mapMechBalances`, and the `paymentData` pass-through). **They all hold.** The kupermind review from 2026-06-08 is strong and its three must-fix items (MPP `closeChannel` ACL, channel⇄mech binding, quote-policy decision) remain open and valid.

The blocker to merge is editorial coherence, not correctness. Two process-level points that aren't line-bound (inline comments below cover the rest):

**PR description is stale/incomplete.** The "Files in this PR" table lists 8 files but omits `docs/mech_prepay_spec.md` — the largest doc (772 lines) and the de-facto v1 recommendation. The summary still frames this as "two candidates, pick one" (x402 OR MPP); there are now **three** proposals.

**The dominant doc is effectively unreviewed.** kupermind reviewed on 2026-06-08; `mech_prepay_spec.md` was rewritten/restructured on 2026-06-09. The newest and most-recommended document has not had a review pass.

Before merge I'd want: (1) one reconciling decision doc naming the single v1 path and relating the other two; (2) PR description updated to list all 9 files and three options; (3) a review pass over `mech_prepay_spec.md`; (4) rebase onto `main` + a 15%-fee / quote-policy consistency sweep (main has since merged proposal_01 activating a 15% fee, PR #161 — every doc here still reasons about 2% / fee=0). Inline nits can ride along.

docs/mpp_session_spec.md
```diff
@@ -0,0 +1,728 @@
...
+    // Escrow does the full validation:
+    //  - decodes paymentData
+    //  - verifies voucher EIP-712 signature
+    //  - requires cumulative > settled
+    //  - transfers (cumulative - settled) USDC to this contract
+    //  - reverts on any tampering or replay
+    mppEscrow.settleFrom(requester, paymentData);
+
+    uint256 received = IERC20(token).balanceOf(address(this)) - before;
+    if (received < deliveryRate) {
```
DavidMinarsch, line 152: **Surplus-settlement branch is undocumented and untested.** If the escrow transfers more than `deliveryRate` (cumulative delta > this batch's summed rates — e.g. the mech under-batches), the override returns `balance + received - deliveryRate`, crediting the surplus to `mapRequesterBalances[requester]`. It nets out against the next batch since `settled` advances by the full delta, but (a) this tracker has no requester withdrawal, and (b) the §10 test list has no case for "settlement delta > summed batch rates." Add a test and a sentence describing where the surplus goes.

docs/x402_key_findings_pre_audit.md
```diff
@@ -0,0 +1,101 @@
...
+
+**Original claim**: Fee can round to zero when `maxDeliveryRate` is very small.
+
+**Why removed**: `_processPayment` in BalanceTrackerBase (line 158) uses **ceil division**:
+```solidity
+marketplaceFee = (balance * fee + (MAX_FEE_FACTOR - 1)) / MAX_FEE_FACTOR;
+```
+When `fee > 0` and `balance >= 1`, `marketplaceFee` is always `>= 1`. Additionally, fees are calculated on accumulated `mapMechBalances[mech]` (after multiple deliveries), not per-request — so small individual delivery rates aggregate before the fee is taken.
+
+## N-3: Multi-Mech Isolation Mechanism Mis-Identified in Spec
```
DavidMinarsch: **N-3 is stale and should be struck through.** It quotes spec §3.7 as saying "Each mech has its own `BalanceTrackerX402` address" and flags it as wrong — but the current `x402_spec.md` §3.7 already describes the shared tracker correctly. As written this reads as an open correction against text that no longer exists; mark it resolved like H-2 / old-M-1 / old-L-1. (Minor: it cites "Issue C" but the spec labels it Issue 2 / Signature 2.)

docs/x402_key_findings_pre_audit.md
```diff
@@ -0,0 +1,101 @@
...
+The x402 integration design is architecturally sound and security-conscious. The key risk is the settlement window between tool execution and on-chain payment (H-1), which the spec explicitly accepts for small amounts. The biggest blocker is Gnosis USDC EIP-3009 verification (M-2).
+
+Three original findings were removed after code review:
+- **H-2** (cross-chain replay): USDC's own domain separator handles this; BalanceTracker doesn't verify signatures.
+- **M-1** (ABI decode safety): Atomic revert on bad `paymentData` is correct behavior; try/catch would allow unpaid deliveries.
+- **L-1** (fee rounding): Ceil division already prevents rounding to zero.
+
+**Critical function to audit post-implementation**: `BalanceTrackerX402._adjustInitialBalance()` — the only new on-chain logic touching user funds.
+
+**Full review**: See [README.md](README.md) for complete analysis including contract checklist, architecture assessment, and recommendations.
```
DavidMinarsch: **Dangling reference.** "Full review: See [README.md](README.md)" — no such README is included in this PR. Either add it or drop the link. Separately, the hard-coded `MechMarketplace` line numbers used throughout these pre-audit docs (232 / 236 / 811) are stale vs current `main` (`getRequestId` is at 883, routing at 857/864). For docs pinned to a branch snapshot, add a "line numbers as of commit `<sha>`" note or refresh them.

CLAUDE.md
```diff
@@ -0,0 +1,89 @@
...
+
+| Model | Mech | Factory | BalanceTracker |
+|-------|------|---------|----------------|
+| Fixed Price Native | `MechFixedPriceNative` | `MechFactoryFixedPriceNative` | `BalanceTrackerFixedPriceNative` |
+| Fixed Price Token | `MechFixedPriceToken` | `MechFactoryFixedPriceToken` | `BalanceTrackerFixedPriceToken` |
+| Nevermined Subscription | `MechNvmSubscription{Native,Token}` | `MechFactoryNvmSubscription{Native,Token}` | `BalanceTrackerNvmSubscription{Native,Token}` |
+
+Chain-specific variants exist under `mechs/native/celo/`, `mechs/token/usdc/`, and `nevermined/token/usdc/`.
+
+**x402 Payment (planned)** — a fourth payment model adding x402 protocol compatibility via EIP-3009 `transferWithAuthorization` for USDC. New contracts (`BalanceTrackerX402`, `MechFixedPriceTokenX402`, `MechFactoryFixedPriceTokenX402`) under `mechs/token/x402/`. Spec: `docs/x402_spec.md`. Implementation plan: `docs/x402_implementation_plan.md`. Key design: overrides `_adjustInitialBalance` in `BalanceTrackerBase` to decode EIP-3009 paymentData instead of using `transferFrom`. Zero changes to MechMarketplace or OlasMech.
```
DavidMinarsch: **Incomplete for what this PR adds.** This section documents only x402 as the "planned fourth payment model." The PR also adds full specs for MPP session (`mpp_session_spec.md`) and mech-prepay (`mech_prepay_spec.md`). Since CLAUDE.md is the assistant-facing project guide, it should mention all three planned families (or at least point to the decision guide), otherwise a code assistant reading this will think x402 is the only direction. (The `0.8.30`/Prague claim is correct.)

docs/x402_spec.md
DIvyaNautiyal07: Renamed in 04f9d4b — now BalanceTrackerX402USDC / MechFixedPriceTokenX402USDC / MechFactoryFixedPriceTokenX402USDC. Kept FixedPrice to match the existing MechFixedPriceTokenUSDC convention and to leave room for an x402 NVM/dynamic variant later. PAYMENT_TYPE is already keccak256("X402USDC") so the names now line up with the routing hash.

docs/x402_spec.md
DIvyaNautiyal07: Applied this in 04f9d4b. Section 3.4 now exposes only POST /verify, with an explicit note on why /supported and /health are omitted: the 402 body's accepts[] already advertises supported config, and the mech's existing healthcheck owns the 503-on-settlement-timeout signal (no new /health endpoint). Section 6.2 defers to §3.4 for the same reason.

docs/x402_spec.md
DIvyaNautiyal07: Same fix as the §3.4 thread — /supported and /health dropped in 04f9d4b. §6.2 now defers to §3.4 for the rationale, and §3.5's /supported response stub is replaced by a pointer to the 402 body's accepts[].

docs/x402_implementation_plan.md
DIvyaNautiyal07: Got addressed earlier in the thread — the original framing in the spec was wrong, batching IS achievable. x402_spec.md §3.3 "Batch semantics" covers it: because paymentData is opaque bytes, BalanceTrackerX402USDC._adjustInitialBalance can decode it as an Authorization[] and loop transferWithAuthorization per element. No MechMarketplace / OlasMech / paymentData-interface changes needed. v1 still recommends per-request settlement for simplicity (atomic-revert isolation), batching is held as a later optimization.

DIvyaNautiyal07: Nothing on the SC level beyond what's already specced. The BalanceTrackerX402USDC._adjustInitialBalance override decodes paymentData as Authorization[] and loops transferWithAuthorization per element — MechMarketplace, OlasMech, and the paymentData interface stay untouched. Off-chain the mech encodes an array per batch instead of a single auth and the off-chain accumulation window has to fit within the shortest validBefore in the batch. See x402_spec.md §3.3 "Batch semantics" for the full description.

docs/x402_vs_mpp.md
DIvyaNautiyal07: Reconciled in 7920d30. The three docs are not three v1 answers — they're three layers: `marketplace_api_spec.md` is the rail (off-chain HTTP request path, structured 402, mech-side write to the data lake, all on the existing token rail with zero new payment contracts), and x402 / MPP are additive payment families a mech operator can later register against `mapPaymentTypeBalanceTrackers`. Updated `x402_vs_mpp.md` §7 "Skip both" to state this explicitly and reference `marketplace_api_spec.md`, and added it as the first entry in §10 References.

docs/x402_spec.md
DIvyaNautiyal07: Rebased fee assumption to 15% (1500 bps) in 7920d30 across x402_spec.md, mpp_session_spec.md, x402scan_integration.md, and x402_vs_mpp.md. Every worked example, JSON snippet, and batch-math line now uses the live fee from governance proposal 01. All examples were re-derived under Policy A (mech absorbs) — see reply on the Policy A/B thread for why I unified rather than parametrized.

docs/x402_spec.md
DIvyaNautiyal07: Added the single-requester constraint to §3.3 "Batch semantics" in 7920d30. The new paragraph cites `MechMarketplace.sol:833-834` and states explicitly that array-encoded paymentData batches one client's requests but cannot cross-settle distinct clients in one tx. Cross-client batching would need a marketplace-side change to accept per-request requester addresses — out of scope for v1.

docs/x402_spec.md
DIvyaNautiyal07: Unified on Policy A in 7920d30. The previous inconsistency was real — §3.3 Policy B formula gave one number (10205 at 2%, 11765 at 15%), §4 / §5 used a different additive form (10200 at 2%). Now every worked example, the §4 routing table, the §5 numbers, and mpp_session_spec.md §5 all use Policy A (`quote = maxDeliveryRate`). I added an explicit note that A and B cannot be mixed inside a single spec because they advertise different `maxAmountRequired` values for the same listed rate. If the team wants Policy B for v1, it's now a clean swap of the quote-derivation function and re-running each example; happy to do that pass if you'd prefer it.

docs/mpp_session_spec.md
DIvyaNautiyal07: Documented the surplus-settlement branch in 7920d30. New paragraph in §3.6 (under the override's Properties block) describes the surplus path: `balance + received - deliveryRate` credits the remainder to `mapRequesterBalances[requester]`; the escrow's `settled` cursor advances by the full delta so the surplus is genuinely paid-for. The next batch from the same requester nets it out because `adjustMechRequesterBalances` debits from the existing balance first. v1 has no requester withdrawal on this tracker, so any surplus that never gets netted is recovered via `MppEscrow.close` (refund path in §5). Added the corresponding test case ("settlement delta > summed batch rates") to the §10 list.

docs/x402_key_findings_pre_audit.md
DIvyaNautiyal07: Struck through N-3 in 7920d30 — marked as resolved with the original text preserved under the heading. Also added a note on the Issue C / Issue 2 / Signature 2 labeling drift you flagged.

docs/x402_key_findings_pre_audit.md
DIvyaNautiyal07: Both fixed in 7920d30. The dangling `README.md` link is gone — replaced with a pointer to `docs/x402_pre_audit.md` which is what "full review" actually meant. For the stale line numbers (232 / 236 / 811 vs current 883 / 857 / 864 on main), I added a commit-SHA pin to the document header: "Line numbers in this document reference MechMarketplace.sol as of commit 6b26aa5 on this branch. They may have drifted on main (e.g. getRequestId is at 883 on main at time of writing). Re-check against the target SHA before quoting." Refreshing every inline citation to current main is a larger sweep — happy to do it if the line-number drift is blocking, otherwise the pin keeps the existing text consistent with itself.

CLAUDE.md
DIvyaNautiyal07: Updated CLAUDE.md in 7920d30 to cover both candidate payment families and to point at the decision guide. The relevant section now lists x402 and MPP session as the two candidates with their respective spec docs, references `docs/x402_vs_mpp.md` as the tiebreaker, and pins the live 15% fee and Policy A quote convention. The off-chain marketplace migration spec is listed separately as the rail (added in an earlier commit on this branch).

docs/x402scan_integration.md
DIvyaNautiyal07: Concrete URL schema proposal: `https://<mech-host>/<chain-id>/<mech-address>/predict` for the payable resource, and `https://<mech-host>/.well-known/x402` for self-discovery (per spec). Examples on Base: `https://mech.olas.example/8453/0xMechAddress/predict` and `https://mech.olas.example/.well-known/x402`. Chain-id in the path makes multi-chain mechs unambiguous (same operator, one per chain), and the mech-address segment lets a single host front more than one mech without route collision. Happy to bake this into the spec if it lands well.

docs/x402scan_integration.md
DIvyaNautiyal07: Agreed — the mech-as-facilitator architecture in `x402_spec.md` §3.4 already covers `/verify` in-process, so the external-facilitator integration in this section is genuinely optional and isn't needed for any marketplace flow. Will tighten this section in a follow-up to make Bazaar via external facilitator strictly an opt-in for operators who want extra ecosystem reach, and to point at the self-verify path as the recommended default. The original framing of two equal paths was misleading.

DavidMinarsch (commented):
## Re-review (head `7920d30`)

Strong response — 7 of my 8 first-pass findings are fully resolved:

- ✅ **Coherent v1 story** — `x402_vs_mpp.md` §7/§10 + CLAUDE.md now frame `marketplace_api_spec.md` as the rail and x402/MPP as additive payment families on top. Clear.
- ✅ **PR description** — now lists all 11 docs and frames MPP as the alternative; migration + analytics specs included.
- ✅ **15% fee sweep** — merged `main`, worked examples re-derived at 1500 bps (the §5 numbers 100000 / 15000 / 85000 check out).
- ✅ **x402 single-requester batching** — §3.3 now states it explicitly with the `MechMarketplace.sol:833-834` citation.
- ✅ **MPP surplus-settlement branch** — documented in §3.2 + §10 test case added.
- ✅ **N-3 struck through**, ✅ **dangling README link** repointed to `x402_pre_audit.md` + commit-SHA pin added, ✅ **CLAUDE.md** lists both families.

**One residual (finding B, quote policy).** The worked examples were converted to Policy A, but the *normative* quote-computation lines were missed and now contradict the new "v1 ships Policy A" statement in §3.3. Four spots still carry the old additive `maxDeliveryRate + (maxDeliveryRate * fee_bps / 10000)` formula (inline comments below). Under Policy A the quote is simply `maxDeliveryRate` — the mech absorbs the fee, it is **not** added on top. Note the additive formula also doesn't match Policy B (`ceil(rate*10000/(10000-fee))`): at 15% on a 10000 rate, additive→11500, Policy A→10000, Policy B→11765, so it's currently a fourth, unintended number.

Fix those four lines and this is consistent end-to-end. Everything else looks good to land.

docs/x402_spec.md
```diff
@@ -0,0 +1,499 @@
...
+---
+
+## 4. End-to-End Flow
+
+All HTTP steps are new. The on-chain batch cycle (steps 9-11) reuses the existing `deliverMarketplaceWithSignatures` path — the only new behavior is inside the overridden `_adjustInitialBalance`.
+
+| Step | Layer | Description |
+|------|-------|-------------|
+| **1** | HTTP — off-chain | Client sends `POST /predict {tool, prompt}` to Mech. No `X-Payment` header present. |
+| **2** | Mech — off-chain | Mech generates a cost estimate. For fixed-price mechs: `quote = maxDeliveryRate + (maxDeliveryRate * fee_bps / 10000)` — no tool execution needed. For NVM/dynamic mechs (future): run tool with `delivery_rate=0` to get cost estimate, then inflate with fee. Queries `MechMarketplace.fee()` for current fee bps. |
```
DavidMinarsch: Residual Policy-B formula — this is the normative "how the mech computes the quote" line, so it matters most. Under the now-canonical Policy A it should read `quote = maxDeliveryRate` (mech absorbs the fee), not `maxDeliveryRate + (maxDeliveryRate * fee_bps / 10000)`. Also drop "then inflate with fee" for the NVM branch — Policy A doesn't inflate. As written this contradicts §3.3 and the §5 worked example (which now quote 10000, not 11500).

docs/x402_spec.md
```diff
@@ -0,0 +1,499 @@
...
+- `processPaymentByMultisig` fee logic: completely unchanged.
+
+### Flow Routing
+
+All four request paths coexist in the mech with no breaking changes to existing flows:
+
+| Condition | Flow |
+|-----------|------|
+| `X-Payment` header present | x402 flow — verify in-process → execute → HTTP 200 |
+| No headers, no `delivery_rate` | x402 quote — return deterministic quote (`maxDeliveryRate + fee`) → HTTP 402 |
```
DavidMinarsch: Same residual: the routing table says `deterministic quote (maxDeliveryRate + fee)`. Under Policy A it's `deterministic quote = maxDeliveryRate`. Update to match §3.3 / §5.

docs/mpp_session_spec.md
```diff
@@ -0,0 +1,730 @@
...
+---
+
+## 4. End-to-End Flow
+
+All HTTP steps are new. The on-chain batch cycle reuses the existing `deliverMarketplaceWithSignatures` path. The only new on-chain behavior is inside the overridden `_adjustInitialBalance` (which calls `MppEscrow.settleFrom`).
+
+| Step | Layer | Description |
+|------|-------|-------------|
+| **1** | HTTP, off-chain | Client sends `POST /predict {tool, prompt}` to Mech. No `Payment-Credential` header. |
+| **2** | Mech, off-chain | Mech generates a cost estimate. For fixed-price mechs: `quote = maxDeliveryRate + (maxDeliveryRate * fee_bps / 10000)`. Queries `MechMarketplace.fee()`. |
```
DavidMinarsch: Same residual in the MPP flow: step 2 still uses `quote = maxDeliveryRate + (maxDeliveryRate * fee_bps / 10000)`. §3.6 and the §5 worked example were moved to Policy A (10000), so this line now contradicts them. Should be `quote = maxDeliveryRate`.

docs/x402_implementation_plan.md
```diff
@@ -0,0 +1,268 @@
...
+
+5. **Registration via `mapPaymentTypeBalanceTrackers`** — correct. A new payment type `keccak256("X402USDC")` maps to the new `BalanceTrackerX402USDC` address, with matching `MechFixedPriceTokenX402USDC` and `MechFactoryFixedPriceTokenX402USDC`.
+
+6. **Mech as Facilitator** — reasonable. Eliminates an external dependency.
+
+7. **Override target `_adjustInitialBalance`** — correctly identified (Section 3.1). The call chain, the need to bypass `_getRequiredFunds`/`transferFrom`, and the `super` fallback for empty `paymentData` are all accurately described.
+
+8. **Three-contract set** — the spec now correctly requires `BalanceTrackerX402USDC`, `MechFixedPriceTokenX402USDC`, and `MechFactoryFixedPriceTokenX402USDC` (Section 3.1 "Required contract set").
+
+9. **Quote generation** — Section 3.3 correctly distinguishes fixed-price (deterministic `maxDeliveryRate + fee`, no tool execution) from NVM/dynamic (future scope).
```
DavidMinarsch: Stale description: "Section 3.3 correctly distinguishes fixed-price (deterministic `maxDeliveryRate + fee` ...)". §3.3 now ships Policy A, where the fixed-price quote is just `maxDeliveryRate`. Re-word so the plan matches the spec it's summarizing.

docs/x402_spec.md
DIvyaNautiyal07: Fixed in 52b5377. §4 step 2 now reads `quote = maxDeliveryRate` (Policy A, mech absorbs the fee). NVM/dynamic branch no longer says "inflate with fee" — Policy A doesn't inflate. The mech still queries `MechMarketplace.fee()` so the `processPaymentByMultisig` carve-out at step 11 is recorded against the live bps, which is why the call is still there.

docs/x402_spec.md
DIvyaNautiyal07: Fixed in 52b5377. Routing table row now reads `deterministic quote (Policy A: quote = maxDeliveryRate)`.

docs/mpp_session_spec.md
DIvyaNautiyal07: Fixed in 52b5377. §3.6 step 2 now reads `quote = maxDeliveryRate` (Policy A), matching §3.6's policy block and the §5 worked example (10000).

docs/x402_implementation_plan.md
DIvyaNautiyal07: Fixed in 52b5377. The §1 item 9 summary now says: "§3.3 ships Policy A: fixed-price quote is `maxDeliveryRate` (mech absorbs the marketplace fee), no tool execution needed. NVM/dynamic-price mechs (future scope) get the same Policy A treatment of the computed estimate (no fee inflation). §3.3 also describes Policy B for completeness but the v1 spec is consistently Policy A across §4, §5, and the MPP session spec." So the plan now matches the spec it's summarizing.

DavidMinarsch (commented):
## Re-review (head `52b5377`) — residual resolved

The last residual (finding B) is fixed. All four normative quote lines now ship Policy A consistently:

- ✅ `x402_spec.md` §4 step 2 → `quote = maxDeliveryRate` (mech absorbs fee)
- ✅ `x402_spec.md` §4 routing table → `Policy A: quote = maxDeliveryRate`
- ✅ `mpp_session_spec.md` §4 step 2 → `quote = maxDeliveryRate`
- ✅ `x402_implementation_plan.md` item 9 → re-worded to Policy A, no fee inflation

Nice touch keeping the `MechMarketplace.fee()` query for the `processPaymentByMultisig` carve-out, and handling the NVM/dynamic branch correctly ("return as-is — Policy A does not inflate").

Full-tree scan confirms **no remaining `maxDeliveryRate + fee` formula and no stale 2% / 200 bps / 10200 fee numbers** anywhere in `docs/` (the one `10200` left in `configuration_v0.json` is a Chiado `chainId`, unrelated to fees).

That clears every first-pass and second-pass finding. From a docs-correctness standpoint this set is now internally consistent and matches the contracts on `main`. No further blocking comments from me — the merge/approve decision is yours.

PR #174, opened 2026-07-06
docs(analytics): analytics ETL spec + metrics schema with instance-grain API and settled mech costs

DIvyaNautiyal07:
## Summary

Reworks the mech analytics ETL spec (`docs/mech_analytics_etl_spec.md`) and the metrics Postgres schema reference (`docs/mech_analytics_etl_schema.md`) so that every field a consumer reads is Olas-public and on-chain-sourced. Applies the on-chain-source rule (metrics come from on-chain when they're on-chain) end-to-end, including for off-chain deployments.

### Sourcing rework

- Per-chain and per-agent mech fees are sourced from **on-chain `MechBalanceAdjusted` events** (indexed by the `new-mech-fees` subgraph extended with a per-`Requester` entity, or, as a fallback, direct RPC `eth_getLogs` on the `BalanceTracker` contracts). Not from the `mech_requests` data lake. The event stream covers on-chain and off-chain requests uniformly because `MechBalanceAdjusted` fires on every mech-balance credit, including the credits emitted during batched off-chain settlement via `deliverMarketplaceWithSignatures`.
- The `mech_requests` data lake is the source only for `per_request_scores` (tool quality — Brier, calibration, per-tool accuracy). It never feeds a public spend or ROI number.

### `agent_aggregates` — all fields public

```sql
n_mech_requests, mech_fee_usd,                    -- from MechBalanceAdjusted per requester
n_bets_omen, n_bets_polymarket,                    -- from Predict subgraph FPMM trades
roi_omen, roi_polymarket,                          -- composed from above two
tool_accuracy_omen, tool_accuracy_polymarket       -- from per_request_scores
```

Every field is on-chain-sourced. No public/internal split, no `_settled` / `_unjoined` variants. Earlier revisions carried a settled/unjoined split for trader's own performance summary, but trader has dropped its ETL dependency (Consumer 4 below) and no other consumer wants a resolved-only pre-aggregate. Add them back additively if needed later.

### API

- Three endpoints unchanged in shape: `GET /v1/metrics/ai-agent/{name}`, `GET /v1/metrics/tool/{name}`, `GET /v1/metrics/ai-agent/{name}/instances`.
- `/instances` serves one entry per Safe with only the public fields above, keyed by `agent_address`. olas-website's ROI distribution chart plots `roi_omen` / `roi_polymarket` directly across instances.
- `/instances?safe_address=` narrows to a single entry; the route validates the passed Safe belongs to the named agent (404 otherwise) so `/ai-agent/trader/instances?safe_address=<optimus-safe>` doesn't leak.

### Consumers (four flows for olas-website, plus two agent-side)

- **Consumer 1a — Marketplace Turnover / Mech page:** unchanged; `new-mech-fees` already covers both request paths.
- **Consumer 1b — Predict-page aggregate ROI:** two breaks under off-chain (mech-subgraph `totalRequests` undercounts, `questionTitle` open-market match fails). Fixed by reading `mech_fee_usd` from `/v1/metrics/ai-agent/{name}`.
- **Consumer 1c — ROI distribution histogram (`roi-distribution.ts`):** full walkthrough of the ~700-line cron-maintained file (5 subgraphs, QMR blob, DEFAULT_MECH_FEE, ROI_BINS). After ETL, the entire file gets deleted; client bins `roi_omen` / `roi_polymarket` from `/instances` client-side.
- **Consumer 1d — Per-tool accuracy KPI:** replaced by `tool_accuracy_omen` on `agent_aggregates`.
- **Consumer 2 — townhall-kpis:** Total Fees from MM unchanged (`new-mech-fees` covers off-chain); per-agent Pearl ROI switches to `/instances`.
- **Consumer 3 — mech-predict:** unchanged — daily report reads `/v1/metrics/ai-agent/{name}`, recompute path reads `per_request_scores` directly via SQL.
- **Consumer 4 — trader agent:** **does not consume the ETL at all.** Keeps the existing on-chain path unchanged (`sender.totalMarketplaceRequests` counter, `parsedRequest.questionTitle` open-market exclusion, hardcoded `DEFAULT_MECH_FEE`). Adds one new term for off-chain pre-deposits, sourced from on-chain `BalanceTracker.Deposit` events via RPC `eth_getLogs`. Zero behaviour change for production on-chain traders (empty log list → term is zero); forward-safe for the off-chain rollout. See §3 Consumer 4 for the full walkthrough with worked examples.
- **Consumer 5 — optimus agent:** no code change needed. `_create_portfolio_data` already books mech pre-deposits as loss the moment funds leave the Safe into the BalanceTracker (contract outflows aren't counted as withdrawals per `_is_not_other_contract_optimism`).

### Peripheral

- New §4.6 input source describes the on-chain event stream and the subgraph-extension-vs-RPC choice.
- §7.10 ROI boundary tightened — every input is explicitly on-chain-verifiable (trade PnL from Predict subgraph, mech cost from `MechBalanceAdjusted`, staking rewards from staking subgraph).
- §7.12 (chain fees) and §7.13 (per-agent) rewritten around the on-chain source.
- Open Question 5 restated: data lake is the source for request **bodies** (needed for `per_request_scores` tool quality metrics), not for public spend numbers.
- New Open Question 6: subgraph-extension vs RPC fallback for the per-requester read path.
- `chain_aggregates` retains its `source` column (`etl_live` / `legacy_snapshot`) with the audited legacy-subgraph decommission snapshot.

## Notes for reviewers

- Every consumer either reads Olas-public on-chain-sourced fields from the ETL, reads its own on-chain source directly (trader), or needs no change (optimus). No consumer reads a Valory-side field for an Olas-facing metric.
- The dropped settled/unjoined variants are an intentional cleanup, not an oversight — the design decision paragraph in §7.13 documents why they're gone and how to add them back additively if a future need appears.
- Open Question 6 is worth a call — do we ship this ETL rollout on the RPC fallback path first, or block on the `new-mech-fees` `Requester` entity extension?

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/mech_analytics_etl_schema.md
```diff
@@ -233,38 +242,69 @@ CREATE TABLE mech_aggregates (
...
-`GET /v1/metrics/agent-economy/{agent_name}` returns a `chain` block keyed by chain name; the route maps `chain_id → name` (gnosis, base, polygon) and reads one row per chain per window.
+`GET /v1/metrics/ai-agent/{ai_agent_name}` returns a `chain` block keyed by chain name; the route maps `chain_id → name` (gnosis, base, polygon) and sums across `source` for the requested window:
+
+```sql
+SELECT chain_id,
+       SUM(total_mech_fees_usd) AS total_mech_fees_usd,
+       SUM(total_requests)      AS total_requests,
+       SUM(total_deliveries)    AS total_deliveries
+  FROM chain_aggregates
+ WHERE window_kind = $1
```
bennyjo, line 311: Is the `all`-window behavior here intended? As I read it, the merge drops the `legacy_snapshot` row for `all`: it's a single frozen row at `window_end = T` (line 281), but the filter pins `window_end = MAX(window_end) WHERE source='etl_live'`, which tracks ~now — so `all`-window chain fees would exclude all legacy history, which seems opposite to the line 281 guarantee ("contributes to every `all`-window read forever"). Rolling windows look fine, since line 282 writes a per-interval snapshot row that shares the live `window_end`. Am I missing something, or should the `all`-window snapshot not be gated on the live max?

bennyjo (commented):
One question on the chain_aggregates legacy-snapshot merge.

docs/mech_analytics_etl_schema.md
DIvyaNautiyal07: Good catch, you're right. The all-window snapshot row is frozen at window_end = T, so the live-max pin would exclude it as soon as the live rows move past T, which contradicts the line 281 guarantee. Rolling windows were fine for the reason you said. Fixed in d504cfb: the filter now has a second branch that includes source='legacy_snapshot' rows unconditionally when window_kind='all', and I added a paragraph under the query explaining why the two branches exist.

bennyjo (commented):
The `all`-window fix in d504cfb looks right — the frozen legacy row is always included for `all`, and no double-counting. Good.

Two small non-blocking notes:

1. Rolling windows have the same frozen-vs-sliding issue you just fixed for `all`: between `T` and `T+window_size`, a rolling window only overlaps part of the legacy period, but the snapshot is one frozen total. Could you add a line saying how the rolling snapshot row is valued, so it's clear those numbers don't over-count?
2. The DDL comment lists `24h` as a window, but it's still an open question and the example JSON leaves it out. Either add it or drop it.

bennyjo (commented):
A few clarifications on the spec/schema consistency and the /instances contract.

docs/mech_analytics_etl_spec.md
```diff
@@ -128,14 +130,17 @@ Today:
 - ROI math in `roi.ts`.
 
 After:
-- One call to `GET /metrics/agent-economy/{agent}` returns mech fees per chain + ROI per agent in the same response.
-- Fee totals come from `chain_aggregates`; per-agent ROI from `agent_aggregates`.
+- One call to `GET /metrics/ai-agent/{agent}` returns mech fees per chain (legacy snapshot + live ETL, merged) and per-agent ROI in the same response.
+- Live fee totals come from `chain_aggregates` rows with `source='etl_live'`. Legacy fee totals come from `chain_aggregates` rows with `source='legacy_snapshot'`, loaded once at legacy-subgraph decommission and never updated again.
```
bennyjo: This line talks about `etl_live` and `legacy_snapshot` rows, but the `chain_aggregates` table definition in §5.5 doesn't have a `source` column at all — it's still the old version, so the spec disagrees with itself. Could you update the §5.5 table to match (add `source`, `snapshot_input_hash`, the new UNIQUE), or replace it with a pointer to the schema doc? The summary table at the top of the schema doc also still shows the old `(chain, window)` grain.

docs/mech_analytics_etl_spec.md
```diff
@@ -682,21 +693,40 @@ total_deliveries    = count of mech_responses with delivered_at in window
...
+### 7.13 Settled mech costs (per agent, per window)
+
+```
+n_mech_requests         = count of mech_requests rows with requester = agent_address and requested_at in window
+mech_fee_usd            = sum(delivery_rate_usd) over those rows that were delivered
+n_mech_requests_settled = subset of the above whose joined market has resolved (resolved_outcome IS NOT NULL)
+mech_fee_usd_settled    = sum(delivery_rate_usd) over the settled subset
+```
+
+Why the settled split exists: trader's performance summary accounts mech costs on market resolution. Today it computes "settled = total - open" by fuzzy-matching its recent requests' `questionTitle` against open Omen markets, which stops working once request content is no longer published to IPFS. The `_settled` fields keep the same accounting semantics through the `market_id` join instead, so trader reads them directly and drops the open-market machinery.
```
bennyjo: This says the `_settled` fields keep "the same" semantics as trader's current math, but there's a gap: trader counts a request as settled once it's *not* matched to an open market, so requests that never match any market (no `market_id`, parse failures) count as settled right away. The new definition leaves those out forever, so their fees never appear in settled costs — a permanent undercount for anyone with unmatched requests. Could you note this difference, or add a count of excluded requests so consumers can see what's missing?

docs/mech_analytics_etl_spec.md
```diff
@@ -720,6 +750,20 @@ Single endpoint. FastAPI in front of the metrics Postgres. No auth on the read s
...
 
+### `GET /metrics/tool/{tool_name}`
+
+One JSON for a single tool with quality metrics that don't depend on the requester: Brier, calibration curve, ECE, BSS, edge, log loss, directional accuracy, no-signal rate. Same windows. Straight read of `tool_aggregates`.
+
+### `GET /metrics/ai-agent/{ai_agent_name}/instances`
+
+An array, one entry per agent instance (Safe) of that service: the same shape as the agent-level response but flattened, so consumers can bin into ROI distribution charts or other cross-instance views client-side. Each entry is keyed by the instance's Safe address, and an optional `safe_address` query param filters the array down to that single entry. This is how a running agent fetches its own numbers, since the Safe address is the only identity an agent knows about itself. Instance entries carry the same fields as the agent-level response, including the `all` window and the settled-only mech cost fields. Trader's Pearl performance summary reads `n_mech_requests_settled` and `mech_fee_usd_settled` from here and keeps composing ROI locally; its trade totals still come from the trader subgraph and staking rewards stay agent-side (§7.10 boundary note).
```
bennyjo: "Same shape as the agent-level response" promises more than the endpoint delivers: the agent-level response is built from three tables (agent, tools, chain), but the schema doc says `/instances` only reads the agent table. Could you list exactly which fields each instance entry contains?

docs/mech_analytics_etl_schema.md
```diff
@@ -202,7 +211,7 @@ SELECT window_end::date AS day,
  ORDER BY window_end ASC;
 ```
 
-`GET /v1/metrics/service/{service_name}` returns an array of these objects, one per agent instance of the service. The route maps `service_name → [agent_address, ...]` and runs the same two queries per agent.
+`GET /v1/metrics/ai-agent/{ai_agent_name}/instances` returns an array of these objects, one per agent instance (Safe) of the service. The route maps `ai_agent_name → [agent_address, ...]` and runs the same two queries per agent. Each entry is keyed by the instance's Safe address. An optional `safe_address` query param narrows the array to that single entry — the same queries with `agent_address = $safe_address` directly, skipping the name-to-addresses fan-out. This is how a running agent fetches its own numbers, since the Safe address is the only identity an agent knows about itself.
```
bennyjo: When `safe_address` is passed, the route looks it up directly and ignores the agent name in the path — so `…/ai-agent/trader/instances?safe_address=<an-optimus-safe>` would return the optimus row under trader's URL. Could you add a line saying the route returns 404 (or empty) when that Safe doesn't belong to the named agent?

docs/mech_analytics_etl_spec.md
```diff
@@ -221,6 +383,27 @@ Per-trade ROI math is in §7.9.
...
+
+| Field | Type | Use |
+|-------|------|-----|
+| `Global.totalFeesInUSD` | BigDecimal | cumulative on-chain fees per chain, backs `chain_aggregates.total_mech_fees_usd` (§7.12) |
+| `Mech.totalFeesInUSD` | BigDecimal | per-mech cumulative fees, backs `mech_aggregates.mech_fee_usd_earned` |
+| `MechBalanceAdjusted` events (log stream, filtered by requester) | — | backs the public per-agent `n_mech_requests` / `mech_fee_usd` in `agent_aggregates` (§7.13) |
+
+The per-requester view (the third row) is not exposed by the current new-mech-fees subgraph, which aggregates per-mech only. Two implementation options:
+
+- **Preferred: extend new-mech-fees with a `Requester` entity** that mirrors the existing `Mech` entity (`totalFeesInUSD` per requester, updated in each `MechBalanceAdjusted` handler using `event.params.requester`). Ships as a small subgraph PR alongside this ETL work. Once shipped, the ETL just queries `Requester(id: safe_address).totalFeesInUSD`.
```
bennyjo: `MechBalanceAdjusted` doesn't carry the requester — its params are `(mech indexed, deliveryRate, balance, rateDiff)` (`contracts/BalanceTrackerBase.sol:50`), so `event.params.requester` doesn't exist and the RPC fallback has no `requester` topic to filter on. `RequesterBalanceAdjusted(requester indexed, deliveryRate, balance)` fires on both request paths and looks like the right source — note `new-mech-fees` doesn't index it today, so the subgraph extension needs a new event handler, not a tweak to the existing one.

docs/mech_analytics_etl_spec.md
```diff
@@ -221,6 +383,27 @@ Per-trade ROI math is in §7.9.
...
+- **Fallback: RPC `eth_getLogs` per BalanceTracker contract**, filter by `requester` topic, sum `deliveryRate`. Runs in the ETL's roll-up job. Slower than a subgraph query, but self-contained if the subgraph extension is not ready.
+
+Both options source from the same on-chain events, so the on-chain-source rule holds either way. The choice is speed vs. subgraph-extension latency. Open questions §15 captures the pick.
+
+**Off-chain requests are covered by the same event stream.** When the mech batches off-chain settlement via `deliverMarketplaceWithSignatures`, the `BalanceTracker` credits each delivered request's fee to the mech balance individually, emitting one `MechBalanceAdjusted` per delivered request. Same handler, same event, same USD conversion — the request's original path is not visible in the event and does not need to be.
```
bennyjo: Both settlement paths emit a single `MechBalanceAdjusted` carrying the batch total, not one per delivered request — `finalizeDeliveryRates` sums `totalMechDeliveryRate` across the batch, and in that path one event can even span multiple requesters. Fee sums still work since the event carries the total, but anything counting events as requests will undercount.

docs/mech_analytics_etl_spec.md
```diff
@@ -672,31 +853,53 @@ This is the field that today's olas-website `tool-accuracy.ts` recomputes from I
...
 
-Sum of `delivery_rate` (in the chain's native currency unit) across `mech_requests.delivered_at` in the window, converted to USD using the chain's stable conversion rate (USDC = $1 for USDC payments, etc.).
+Sourced from the on-chain `MechBalanceAdjusted` event stream indexed by the new-mech-fees subgraph, aggregated per chain per window. Every credit to a mech's balance in the `BalanceTracker` contract emits a `MechBalanceAdjusted(mech, deliveryRate, balance, rateDiff)` event, regardless of whether the request that produced the delivery was on-chain (per-request settlement) or off-chain (batched settlement via `deliverMarketplaceWithSignatures`). The fee stream therefore already covers both request paths.
 
 ```
-total_mech_fees_usd = sum(delivery_rate_usd) over mech_requests delivered in window
-total_requests      = count of mech_requests with requested_at in window
-total_deliveries    = count of mech_responses with delivered_at in window
+total_mech_fees_usd = sum(MechBalanceAdjusted.deliveryRate converted to USD) in window
+total_requests      = count of MechBalanceAdjusted events in window       # 1 event per delivered request
```
bennyjo: Since `MechBalanceAdjusted` fires once per settlement batch, `total_requests = count of events` counts batches, not requests — reintroducing the same undercount shape this spec sets out to fix. Request/delivery counts are available from the marketplace's own requester-indexed events (`MarketplaceRequest.numRequests`, `MarketplaceDelivery.numDeliveries`, `MarketplaceDeliveryWithSignatures.numDeliveries`).

docs/mech_analytics_etl_spec.md
```diff
@@ -672,31 +853,53 @@ This is the field that today's olas-website `tool-accuracy.ts` recomputes from I
...
+**Rule the numbers respect.** These fields are Olas-public. Their source (on-chain events) is verifiable by anyone with an RPC endpoint or a subgraph client, regardless of whether the underlying request originated on-chain or off-chain. The ETL only aggregates; it does not synthesize new fee information from the `mech_requests` data lake.
+
+**What if the new-mech-fees subgraph does not yet index by requester.** For the per-chain aggregate this doesn't matter — chain-level sums are what the existing subgraph already exposes (`Global.totalFeesInUSD`). For the per-agent version (§7.13), the ETL either extends the subgraph with a per-`Requester` entity or falls back to reading `MechBalanceAdjusted` logs directly via RPC and aggregating on write. See "Open questions" §15 for the pick.
+
+### 7.13 Per-agent mech spend
+
+Public Olas-facing fields for per-agent mech cost, sourced from on-chain `MechBalanceAdjusted` events indexed by requester.
+
+```
+n_mech_requests   = count of MechBalanceAdjusted events with requester = agent_address in window
```
bennyjo: Same two issues as §4.6/§7.12 apply here: no `requester` on `MechBalanceAdjusted` to filter by, and event count ≠ request count. If this moves to `RequesterBalanceAdjusted`, note it fires at debit time at the agreed rate (overpayment refunds are silent), so per-agent spend won't exactly reconcile with §7.12's credit-side totals — worth pinning which semantics "spend" means. The mirrored text in `docs/mech_analytics_etl_schema.md` ("Computation sources", ~line 183) needs the same fix.

docs/mech_analytics_etl_spec.md
```diff
@@ -102,48 +104,208 @@ One pipeline. Each consumer reads pre-computed JSON. No subgraph or IPFS calls a
...
+
+Off-chain trader that has pre-deposited 3 times ($1, $2, $2), used only 200 of ~500 requests worth:
+- `settled_mech_requests = 0 - 0 = 0` (no MarketplaceRequest events fired)
+- `onchain_mech_cost = $0`
+- `offchain_prepaid_sum = $1 + $2 + $2 = $5`
+- `total_costs = totalTradedSettled + totalFeesSettled + $0 + $5`
+- Money truly committed ($5), reflected as loss immediately per the pre-deposit-as-loss rule. If the trader wants a smoother ROI curve, reduce `offchain_deposit_target_calls` in `mech_marketplace_config` so top-ups are smaller and more frequent.
+
+Mixed history:
+- Both terms add cleanly. Formula makes no branch on `use_offchain`.
```
bennyjo, line 281: Is "both terms add cleanly" guaranteed? If an on-chain request ever draws down a pre-deposited `BalanceTracker` balance instead of attaching payment, that spend counts twice (once via the `Deposit` sum, once via `settled_mech_requests * 0.01`). Happy to leave as-is if trader always attaches payment per on-chain request — just worth stating that assumption here.

docs/mech_analytics_etl_spec.md
```diff
@@ -672,31 +853,53 @@ This is the field that today's olas-website `tool-accuracy.ts` recomputes from I
...
 
 ## 8. Wildcard API
 
-Single endpoint. FastAPI in front of the metrics Postgres. No auth on the read side; Cloudflare in front.
+Three endpoints. FastAPI in front of the metrics Postgres. No auth on the read side; Cloudflare in front. They match the three natural axes of analysis: one agent's economy, one tool's quality, one service's instance distribution. Each is extensible by adding keys to its JSON, never by adding new routes. Today's olas-website "verify" links become curl invocations of these endpoints, preserving the verifiability pattern.
+
+### `GET /v1/metrics/ai-agent/{ai_agent_name}`
 
-`GET /metrics/agent-economy/{agent_name}` returns:
+One JSON containing the agent's mech-request counts and fees (totals and settled-only variants), ROI per platform, tool accuracy per platform, the per-tool metric breakdown, and per-chain totals. Rolling windows (7d / 30d / all) plus per-day snapshots for ROI and tool accuracy.
```
bennyjo: Leftover from dropping the `_settled`/`_unjoined` fields: this line still promises "settled-only variants", and `docs/mech_analytics_etl_schema.md:218` similarly still says `/instances` entries include "the settled and unjoined variants" — both need scrubbing to match the removed columns.

bennyjo (commented):
Findings on the new on-chain event-sourcing sections (§4.6/§7.12/§7.13), the trader Deposit-events design, and two leftovers from dropping the settled/unjoined fields.

docs/mech_analytics_etl_spec.md
```diff
@@ -927,11 +1154,9 @@ Small list. All non-blocking for starting code.
...
+   - Option A (rejected): ETL pulls onchain requests from the marketplace subgraph periodically and normalizes them into the same shape. Kept the data lake offchain-only and left a permanent live subgraph dependency inside the ETL for the internal fields.
+6. Per-requester mech fees — subgraph extension or RPC? Two ways to source the public per-agent mech-spend fields (§4.6, §7.13). Recommendation: extend the `new-mech-fees` subgraph with a `Requester` entity mirroring the existing `Mech` entity. Ship as a small companion PR to autonolas-subgraph-studio ahead of ETL rollout. RPC fallback stays available if the subgraph extension slips.
```
bennyjo: One consideration for this pick: `agent_aggregates` needs 7d/30d windows and daily snapshots, but a `Requester` entity mirroring `Mech` is a cumulative lifetime total — it can only serve the `all` window. The extension needs per-requester event-level or per-day granularity (e.g. a requester field on `MechTransaction`, or a `RequesterDaily`), or the RPC log-scan becomes the primary path — either way the companion subgraph PR is bigger than a mirror of the `Mech` entity.

docs/mech_analytics_etl_spec.md
```diff
@@ -653,6 +830,10 @@ Trades for a window are only counted if their FPMM is resolved within that windo
 
 Where the trade-to-mech-request join happens: by `fpmm.id == market_id` from the mech request's raw_content. This is the join that replaces the fuzzy `questionTitle` match olas-website and townhall-kpis do today.
 
+**Boundary: trading ROI only.** This metric is bet cost vs bet payout. It deliberately excludes mech fees (reported separately as `mech_fee_usd`, §7.13), OLAS staking rewards, and any token price conversion of rewards. Consumers that display a rewards-inclusive ROI (olas-website's `finalRoi`, trader's in-agent `final_roi`) compose it on their side: trading components from this API, mech cost from `agent_aggregates.mech_fee_usd` (also on-chain-sourced, see §7.13), staking rewards from the staking subgraph, and an OLAS/USD price from their own source. Staking data is deliberately out of ETL scope (§16); a precomputed ratio cannot have rewards added to its numerator afterwards, which is why the API also exposes the underlying components rather than only the ratio.
```
bennyjo, line 850: This says the API "exposes the underlying components rather than only the ratio", and §1c promises `/instances` entries carry "the underlying trade components" — but `agent_aggregates` has no cost/payout columns, only the finished `roi_*` ratios. Without e.g. `total_cost_usd` / `total_payout_usd` per platform per window, consumers can't compose `partialRoi`/`finalRoi` as described, and plotting `roi_omen` directly changes the histogram's metric definition (today's denominator includes mech fees).

docs/mech_analytics_etl_spec.md
```diff
@@ -720,6 +923,30 @@ Single endpoint. FastAPI in front of the metrics Postgres. No auth on the read s
...
+
+- `agent_address` (the Safe key) and `agent_name`
+- `windows`: `7d`, `30d`, `all` — each with `n_mech_requests`, `mech_fee_usd`, `n_bets_omen`, `n_bets_polymarket`, `roi_omen`, `roi_polymarket`, `tool_accuracy_omen`, `tool_accuracy_polymarket`. All fields are Olas-public and on-chain-sourced (see §7.13, §7.10).
+- `daily`: the same per-day snapshot series the agent-level response exposes
+
+Consumers rendering the ROI distribution chart on olas-website plot `roi_omen` (or `roi_polymarket`) directly, one dot per Safe, into ROI bins client-side.
+
+Each entry is keyed by the instance's Safe address. An optional `safe_address` query param narrows the array to that single entry. The route validates that the passed Safe belongs to the named agent via the same static name-to-Safes mapping the agent-level route uses; a `safe_address` that does not map to `{ai_agent_name}` returns HTTP 404 (empty body). This prevents `…/ai-agent/trader/instances?safe_address=<an-optimus-safe>` from returning optimus data under the trader path. This is how a running agent fetches its own numbers, once its config gives it both its agent name and its Safe address.
```
bennyjo: Today the ROI histogram discovers Safes dynamically from the Predict subgraph's `traderAgents`, so new Pearl agents appear automatically — a static Safe-to-name file can't do that, and `/instances` would silently miss every newly-created Safe. Suggest a dynamic discovery source (Predict subgraph or service registry) for instance enumeration, keeping the static file for naming only.

bennyjo (commented):
Follow-up architecture findings: windowing vs the cumulative Requester entity, missing ROI components, and instance discovery. One note on text outside this diff: §6.1's cursor filters on `requested_at`, but §13's failure table says late rows are safe because "the cursor is on `created_at`, not `requested_at`" — one of the two needs updating, since under §6.1 as written a late-inserted row with an earlier `requested_at` is skipped forever.

docs/mech_analytics_etl_schema.md
DIvyaNautiyal07: Yes, that's the intended behaviour. The reply above and d504cfb captured the fix — the merge now has a second branch that includes `source='legacy_snapshot'` rows unconditionally on the `all` window.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: The §5.5 spec block is just a pointer to `docs/mech_analytics_etl_schema.md` §7 to avoid drift, and the schema doc's summary table already shows `(chain, window, source)` and its §7 DDL carries `source`, `snapshot_input_hash`, and the UNIQUE — so the on-disk state is consistent. If you were looking at an earlier commit that still had the old grain in either place, refresh from HEAD and it should be clean now.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: The `_settled` / `_unjoined` variants are gone from the schema entirely — see the paragraph in spec §7.13 ('No settled/unjoined split in the schema') and the mirror in schema.md §5 ('Computation sources'). Trader dropped its ETL dependency (spec Consumer 4) and no other consumer needed the split. So the unmatched-request undercount you flagged doesn't actually ship. If a future consumer needs a resolved-only view, we'd add it additively per §8, and this pre-condition on parse failures / non-market requests is a good thing to have documented there when it does — I'll paste it in the follow-up.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Fixed in 4479a3e. Replaced 'same shape as the agent-level response' with an explicit field list on the `/instances` entries: `n_mech_requests`, `mech_fee_usd`, `roi_omen`, `roi_polymarket`, `tool_accuracy_omen`, `tool_accuracy_polymarket`, `n_bets_omen`, `n_bets_polymarket` per rolling window, plus the per-day snapshots. The per-tool and per-chain composition blocks explicitly do NOT appear per-instance since those tables aren't per-Safe.

docs/mech_analytics_etl_schema.md
DIvyaNautiyal07: This is already covered on schema.md line 220 (right after the `/instances` field-shape paragraph): 'The route validates that the passed Safe belongs to the named agent by checking the same static mapping; a `safe_address` that does not map to `{ai_agent_name}` returns HTTP 404 (empty response body). This prevents `…/ai-agent/trader/instances?safe_address=<an-optimus-safe>` from returning optimus data under the trader path.' Let me know if you'd rather have that line duplicated somewhere else in the spec doc too.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Confirmed on both points and fixed the sourcing in 4479a3e. Per-Safe counts now come from the marketplace subgraph's `sender` entity (`sender.totalMarketplaceRequests` bumped by `handleMarketplaceRequest`, `sender.totalLegacyRequests` bumped by `handleMarketplaceDeliveryWithSignatures`), both indexed by requester, covering both paths. This drops the whole `MechBalanceAdjusted per-requester` framing that never worked. Fees on the per-Safe field stay as `n_mech_requests × DEFAULT_MECH_FEE_USD` for now — same formula olas-website ships today. Exact per-Safe sums via `RequesterBalanceAdjusted` (which as you flagged, isn't indexed in `new-mech-fees` today) become the follow-up when dynamic per-request pricing lands. Called that out explicitly in §4.6 and §15.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: You're right. §7.12 now says explicitly that `MechBalanceAdjusted` fires once per settlement action carrying the batch total, not once per delivered request, and that sums are fine but event counts don't equal request counts. Request/delivery counts on that section switched to the marketplace's requester-indexed events (`MarketplaceRequest.numRequests`, `MarketplaceDelivery.numDeliveries`, `MarketplaceDeliveryWithSignatures.numDeliveries`) so we don't reintroduce the undercount you flagged in the neighbouring comment. Commit 4479a3e.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Fixed in 4479a3e. `total_requests` and `total_deliveries` in §7.12 now source from the marketplace's own requester-indexed events (`MarketplaceRequest.numRequests`, `MarketplaceDelivery.numDeliveries` + `MarketplaceDeliveryWithSignatures.numDeliveries`) instead of counting `MechBalanceAdjusted` events. Same reasoning applies to §7.13 which switched to per-Safe counters on the same events — no undercount from batched settlement.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Fixed in 4479a3e. §7.13 now sources per-Safe counts from the marketplace subgraph (§4.6) with `mech_fee_usd = n_mech_requests × DEFAULT_MECH_FEE_USD`, and the mirror in `schema.md` line 183 ('Computation sources' paragraph under `agent_aggregates`) is updated to match. Called out that exact fee sums via `RequesterBalanceAdjusted` (with its debit-time semantics and silent refund-on-overpay behaviour) are the follow-up when the marketplace transitions to dynamic pricing.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Good call. Added the assumption explicitly to trader Consumer 4 in 4479a3e: today on-chain requests attach payment per request via `MechMarketplace.request`, not by debiting a pre-deposited BalanceTracker balance, which is why the `Deposit` sum and `settled_mech_requests × DEFAULT_MECH_FEE` add cleanly without double-counting. Also named the fix path if that ever changes — swap the two source-specific terms for one `RequesterBalanceAdjusted` sum.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Scrubbed in 4479a3e. The `/v1/metrics/ai-agent/{ai_agent_name}` blurb in the spec (line 900) no longer promises 'totals and settled-only variants', and the `/instances` paragraph in `schema.md` (line 218) drops the 'settled and unjoined variants' phrase and gives an explicit field list instead.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Fair point on the granularity, and it made me revisit the whole 'extend the subgraph' framing. That preference was mine as spec author, no subgraph-team commitment behind it. Given `RequesterBalanceAdjusted` isn't indexed in `new-mech-fees` today and rolling windows would need per-day granularity anyway, the subgraph companion isn't a small mirror. So the spec now says: per-Safe counts come from the marketplace subgraph's existing `sender` entity (both paths covered, indexed by requester, no extension), and `mech_fee_usd = n_mech_requests × DEFAULT_MECH_FEE_USD` matches what olas-website ships today. Exact per-Safe fee sums move to a follow-up (indexed `RequesterBalanceAdjusted` OR ETL-side RPC log-scan) when dynamic per-request pricing lands. Commit 4479a3e.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Real gap. `agent_aggregates` today only has finished `roi_*` ratios per window, so consumers can't recompose partial or final ROI on their side. This is a schema decision I'd like to align with you before writing it in: two shapes on the table are (a) add `total_cost_usd` and `total_payout_usd` per platform per window to `agent_aggregates`, keeping the finished ratios for cheap direct reads, or (b) narrow the API contract to 'ratio only, no client-side recomposition, use the /per_request_scores drilldown if you need components'. Preference?

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Also open. A static Safe-to-name file can't discover new Pearl agents, agreed. Three options and I'd like your call before rewriting: (a) subscribe to the Predict subgraph's `traderAgents` (dynamic, but Predict-specific, misses non-Predict agents), (b) enumerate from the on-chain service registry (path-agnostic, catches every registered agent, adds an RPC call per rollup), (c) keep the static file for naming only and layer a dynamic discovery source on top for `/instances` enumeration. I lean (c) since it separates 'what to enumerate' from 'what to call it in the UI', but happy with any of the three.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Went with option (a) in 85f1297. Added `total_cost_omen_usd`, `total_payout_omen_usd`, `total_cost_polymarket_usd`, `total_payout_polymarket_usd` per rolling window per Safe to `agent_aggregates`. Sourced from the Predict subgraphs' FPMM trades entity, restricted to resolved markets, mirroring the `totalTradedSettled + totalFeesSettled` shape trader uses for cost and the realized payout on resolved bets. Kept the finished `roi_*` ratios for cheap direct-ratio reads. Consumers rendering `finalRoi` now recompose it per Safe from `total_cost_*_usd`, `total_payout_*_usd`, `mech_fee_usd`, plus their own staking-subgraph read and OLAS/USD price feed, without a `/per_request_scores` drilldown.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Went with option (b) in 85f1297, closer to what you suggested. The `agent_name → service_id` mapping stays as a small static naming file (that's just the display label), and the `service_id → [safe_address, ...]` step reads from the on-chain ServiceRegistry contract at rollup time and again at each `/instances` request. Newly-registered Pearl instances appear on the next request without an ETL config change. Cross-tenant `safe_address` queries stay guarded via a ServiceRegistry membership check against the resolved `service_id`, so if a Safe re-registers under a different service later the 404 boundary still holds. §15 open question 2 marked resolved.

docs/mech_analytics_etl_spec.md
```diff
@@ -672,31 +865,59 @@ This is the field that today's olas-website `tool-accuracy.ts` recomputes from I
...
+**Rule the numbers respect.** These fields are Olas-public. Their source (on-chain events) is verifiable by anyone with an RPC endpoint or a subgraph client, regardless of whether the underlying request originated on-chain or off-chain. The ETL only aggregates; it does not synthesize new fee information from the `mech_requests` data lake.
+
+**Chain-level sums vs per-Safe attribution.** Chain-level sums are what `new-mech-fees` already exposes (`Global.totalFeesInUSD`), so §7.12 needs no subgraph extension. Per-Safe attribution (§7.13) reads from a different source: the marketplace subgraph's per-Safe request counters (§4.6). Chain-level fees and per-Safe fees are sourced independently by design, so no cross-source reconciliation is needed here.
+
+### 7.13 Per-agent mech spend
+
+Public Olas-facing fields for per-agent mech cost, sourced from the marketplace subgraph's per-Safe request counters and the marketplace's per-request fee constant.
+
+```
+n_mech_requests   = sender.totalMarketplaceRequests + sender.totalLegacyRequests
```
bennyjo: This sum double-counts on-chain requests: `handleMarketplaceRequest` bumps **both** counters (`mech-marketplace.ts:427-428` — `totalMarketplaceRequests += 1`, `totalLegacyRequests += numRequests`), which is the same senderTotal double-count §1c calls a pre-existing bug. `sender.totalLegacyRequests` alone is the both-paths counter (bumped by `numRequests` on-chain and `numDeliveries` off-chain) — same fix applies to §4.6 and the schema doc's Computation sources.

docs/mech_analytics_etl_spec.md
```diff
@@ -310,20 +499,33 @@ CREATE TABLE agent_aggregates (
     window_kind             TEXT        NOT NULL,
     window_start            TIMESTAMPTZ NOT NULL,
     window_end              TIMESTAMPTZ NOT NULL,
-    n_mech_requests         INTEGER     NOT NULL,
-    mech_fee_usd            DOUBLE PRECISION,
-    n_bets_omen             INTEGER,
+    -- All fields below are Olas-public and on-chain-sourced.
+    n_mech_requests         INTEGER     NOT NULL,       -- delta of sender.totalMarketplaceRequests + totalLegacyRequests over the window
+    mech_fee_usd            DOUBLE PRECISION,           -- sum of deliveryRate for those events, converted to USD
```
bennyjo: Stale comment from the previous sourcing: `mech_fee_usd` is now `n_mech_requests × DEFAULT_MECH_FEE_USD`, not "sum of deliveryRate for those events" — same line in `docs/mech_analytics_etl_schema.md:162`.

docs/mech_analytics_etl_spec.md
```diff
@@ -310,20 +499,33 @@ CREATE TABLE agent_aggregates (
     window_kind             TEXT        NOT NULL,
     window_start            TIMESTAMPTZ NOT NULL,
     window_end              TIMESTAMPTZ NOT NULL,
-    n_mech_requests         INTEGER     NOT NULL,
-    mech_fee_usd            DOUBLE PRECISION,
-    n_bets_omen             INTEGER,
+    -- All fields below are Olas-public and on-chain-sourced.
+    n_mech_requests         INTEGER     NOT NULL,       -- delta of sender.totalMarketplaceRequests + totalLegacyRequests over the window
```
bennyjo: The sender counters are cumulative lifetime totals — how does the ETL compute window deltas? It needs either periodic counter snapshots persisted ETL-side or block-height subgraph queries, and backfilling windows before ETL launch only works with the block-height approach. Worth a sentence in §4.6 or §6 naming the mechanism.

docs/mech_analytics_etl_spec.md
```diff
@@ -720,6 +941,30 @@ Single endpoint. FastAPI in front of the metrics Postgres. No auth on the read s
...
+
+**Instance enumeration is dynamic.** The route resolves `ai_agent_name → service_id` from the static naming mapping and enumerates the current Safes for that service from the on-chain ServiceRegistry contract, so newly-registered Pearl agents appear in the array on the next request without an ETL config change. Each entry is keyed by the instance's Safe address. An optional `safe_address` query param narrows the array to that single entry. The route validates that the passed Safe currently belongs to the named service by looking it up in the ServiceRegistry against the resolved `service_id`; a `safe_address` that does not currently belong to `{ai_agent_name}`'s service returns HTTP 404 (empty body). This prevents `…/ai-agent/trader/instances?safe_address=<an-optimus-safe>` from returning optimus data under the trader path, even if the Safe was previously registered under a different service. This is how a running agent fetches its own numbers, once its config gives it both its agent name and its Safe address.
```
bennyjo: Does `service_id` here account for Pearl's one-service-per-user model? Each Pearl user mints their own service registration, so "trader" is thousands of service_ids each with one multisig — a single `service_id` lookup returns one Safe, not the fleet the ROI histogram needs. The static mapping likely needs to resolve to an agent/blueprint ID and enumerate all registered services using it.

bennyjo (commented):
Round-3 findings on the new sender-counter sourcing: a double-count in the §7.13 formula, stale DDL comments, the unspecified window-delta mechanism, and a question on service_id vs Pearl's one-service-per-user model.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Good catch, real bug in the sourcing. Fixed in 7c29ab3. Dropped the sum across §4.6, §7.13, and the schema doc's `agent_aggregates` computation-sources paragraph — `sender.totalLegacyRequests` alone is the both-paths counter (`+= numRequests` on the on-chain path via `handleMarketplaceRequest`, `+= numDeliveries` on the off-chain path via `handleMarketplaceDeliveryWithSignatures`). Adding `totalMarketplaceRequests` on top would over-count every on-chain request, matching the pre-existing double-count I already called out at spec §7.2 line 164 without connecting it back to my own sourcing. Named the counter shape and the reason for not summing directly in the spec text so this stays sticky.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Fixed in 7c29ab3. Both the spec §5.3 column comment and the mirror at schema.md line 162 now read `n_mech_requests * DEFAULT_MECH_FEE_USD (matches olas-website ROI histogram formula; approximation until dynamic per-request pricing lands)`, matching the actual sourcing.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Real gap in the spec, addressed in 7c29ab3. Added a paragraph in §4.6 naming the mechanism: `sender.totalLegacyRequests` is cumulative, so window deltas come from either (a) persisted per-Safe snapshots in `cursor_state` at rollup time (steady state), or (b) block-height subgraph queries `sender(id: safe, block: {number: N}) { totalLegacyRequests }` for backfill of history before ETL launch. Steady-state can use either mechanism; snapshots are cheaper because they avoid per-window block-height queries at every tick. Called out that backfill only works with the block-height path.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: You're right, my enumeration model was wrong for Pearl's shape. Fixed in 7c29ab3. `agent_name` now maps to an `agent_id` (AgentRegistry blueprint entry — e.g. "trader"), not a `service_id`. Pearl mints one service_id per user, so the correct dynamic step is `agent_id → [service_id, ...] → each service's canonical multisig`, all read from the on-chain ServiceRegistry at rollup time. Updated §7.11, §7.13, the milestone checklist, §15 open question 2, and the schema doc's `/instances` route description. The `safe_address` membership guard moves to an agent_id-level check accordingly (a Safe belongs to `{ai_agent_name}` if any currently-registered service under the resolved `agent_id` owns it), so cross-blueprint queries like `trader/instances?safe_address=<optimus-safe>` still 404 correctly.

kupermind (dismissed):

docs/mech_analytics_etl_spec.md
```diff
@@ -221,6 +385,35 @@ Per-trade ROI math is in §7.9.
...
+
+Do **not** add `sender.totalMarketplaceRequests` on top of that. `handleMarketplaceRequest` bumps both counters at once (a pre-existing double-count called out at §7.2 line 164), so summing them would over-count every on-chain request. `totalLegacyRequests` alone is the correct both-paths count.
+
+**Window deltas from a lifetime counter.** `sender.totalLegacyRequests` is cumulative, so per-window numbers come from a delta against a lower bound:
+
+- Steady-state rolling windows (7d / 30d): the ETL persists a per-Safe snapshot `(safe, counter_at_snapshot_ts)` at the leading edge of each window during the rollup job. Window delta = current counter at `T` minus snapshot at `T - window_size`. Snapshots live in `cursor_state` alongside the roll-up watermarks.
```
bennyjo: Small schema wrinkle: `cursor_state` is one row per job with fixed job-level columns (`job_name` PK, §5.6), so it can't hold a per-Safe snapshot series — at Pearl scale that's thousands of `(safe, counter, ts)` rows per window boundary. Probably wants its own small table (or lean on the block-height query path for steady state too).

docs/mech_analytics_etl_spec.md
```diff
@@ -720,6 +944,30 @@ Single endpoint. FastAPI in front of the metrics Postgres. No auth on the read s
...
...s the AgentRegistry blueprint entry — Pearl mints one `service_id` per user, so an agent blueprint like `trader` has thousands of registered service_ids each with a single multisig, not one service_id with a fleet). It then enumerates the current registered services against that `agent_id` from the on-chain ServiceRegistry contract and collects each service's canonical multisig. Newly-registered Pearl agents appear in the array on the next request without an ETL config change. Each entry is keyed by the instance's Safe address. An optional `safe_address` query param narrows the array to that single entry. The route validates that the Safe belongs to a currently-registered service under the same `agent_id` by looking it up in the ServiceRegistry; a `safe_address` that does not currently belong to any service under `{ai_agent_name}`'s `agent_id` returns HTTP 404 (empty body). This prevents `…/ai-agent/trader/instances?safe_address=<an-optimus-safe>` from returning optimus data under the trader path, even if the Safe was previously registered under a different agent. This is how a running agent fetches its own numbers, once its config gives it both its agent name and its Safe address.
```
bennyjo: "Enumerates the current registered services against that `agent_id`" — the ServiceRegistry has no agent_id → services reverse index on-chain, so this is either a full serviceId scan via multicall or a registry-subgraph query. Worth naming which, since it runs on every rollup cycle and `/instances` request.

docs/mech_analytics_etl_spec.md
```diff
@@ -102,48 +104,210 @@ One pipeline. Each consumer reads pre-computed JSON. No subgraph or IPFS calls a
...
+
+Returns an array — one entry per omenstrat Safe — with `windows: { "7d" | "30d" | "all" }` each carrying `roi_omen`, `n_bets_omen`, `mech_fee_usd`, and the underlying trade components. Client-side, olas-website:
+
+1. Deletes `roi-distribution.ts` in its entirety (cron, blob, QMR normalization, TTL flush, byDay pruning, questionTitle fuzzy matcher, DEFAULT_MECH_FEE constant, sender-total double-count bug — all of it).
+2. For each entry, bins `roi_omen` for the selected window into `ROI_BINS` client-side and renders.
+3. Applies the same `MIN_TRADES_FOR_ROI_DISPLAY` guardrail against `n_bets_omen`, which is served on the same entry.
+
+Every field the histogram needs is on-chain-sourced by the ETL: `mech_fee_usd` computed from the marketplace subgraph's per-Safe request counters (§4.6) covering on-chain and off-chain paths uniformly, `roi_omen` composed from Omen Predict subgraph trades and the same on-chain mech cost. No QMR, no `questionTitle` matching, no 90-day retention window logic. Off-chain undercount in the 7d / 30d / 90d tabs disappears as a side effect.
+
+For polystrat the same shape applies with the Polymarket Predict subgraph and the polygon BalanceTracker events.
```
bennyjo: Leftover from the sourcing switch: per-Safe counts now come from the marketplace subgraph's `sender` counters (§4.6), so this should read "polygon marketplace subgraph counters" rather than "polygon BalanceTracker events".

bennyjo (commented):
Three small residuals from the round-4 fixes — all polish, nothing blocking; the sender-counter sourcing and blueprint enumeration both verify correct.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Fixed in d7e579d. Line 183 now reads "polygon marketplace subgraph's per-Safe request counters" to match the sourcing switch that landed in 7c29ab3.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Right — cursor_state can't carry the (safe, counter, ts) series. Reworked the window-delta mechanism (§4.6) in d7e579d: primary path is block-height subgraph queries per Safe at each rollup tick for both steady state and backfill (same mechanism, no snapshot table needed). If block-height queries prove too expensive at Pearl scale (thousands of Safes × per-tick subgraph reads bumping into rate limits), the fallback is a dedicated `sender_snapshot(safe, counter, snapshot_ts)` table. Called out explicitly rather than left implicit.

docs/mech_analytics_etl_spec.md
DIvyaNautiyal07: Named it in d7e579d. §7.13 now says the agent_id → services enumeration reads from the `autonolas-registries` subgraph, which maintains the reverse index. Also called out why (ServiceRegistry on-chain has no such index, so a raw contract enumeration would require a full serviceId multicall per rollup cycle and per /instances request; the subgraph collapses that to a single indexed query). Safe-belongs-to-service validation on the query-param path uses the same subgraph.

bennyjo (approved):
All findings from the five review rounds are addressed and verified — the sourcing chain checks out end-to-end against the BalanceTracker/MechMarketplace contracts, the marketplace subgraph handlers, and the new-mech-fees and service-registry schemas. Remaining trade-offs (fee approximation, deferred exact-fee follow-up) are deliberate and documented. LGTM.

kupermind (approved):

PR #17, opened 2023-11-03
chore: messy fix to static request id requirement

DavidMinarsch:
Sorry for this rather messy fix. It means the trader does not need to change anything.

PR #175, opened 2026-07-06
docs: document tracker fundability after mech factory disabling

kupermind:
## Summary

Adds vulnerability entry **#17 (Informative)** to `docs/Vulnerabilities_marketplace.md`: balance trackers remain fundable after all corresponding mech factories are disabled.

- De-whitelisting a branch's mech factories via `setMechFactoryStatuses` blocks new mech creation; with no mech of that branch, no request can route to its balance tracker.
- Balance trackers are immutable, permissionless contracts: `deposit()` / `depositFor()` and raw token transfers always remain possible, so a zero tracker balance cannot be guaranteed by any governance action.
- Such funds can never be processed as payment (no request path, so `checkAndRecordDeliveryRates`, `finalizeDeliveryRates`, `processPayment` are unreachable) and remain withdrawable by the depositor.
- Frames tracker immutability as a deliberate trustlessness property: no owner, governance action, or upgrade can seize, freeze, or redirect requester deposits, and `drain()` can only ever move `collectedFees`.

### DIFF
--- a/docs/Vulnerabilities_marketplace.md
+++ b/docs/Vulnerabilities_marketplace.md
@@ -22,6 +22,7 @@
 | 14 | `trackerBalance` not decremented on withdrawal | BalanceTrackerNvmSubscriptionNative | Informative |
 | 15 | `SubscriptionProvider.fulfill()` permissionless | SubscriptionProvider | Informative |
 | 16 | `BalanceTrackerNvmSubscriptionNative.depositFor()` allows direct deposits | BalanceTrackerNvmSubscriptionNative | Low |
+| 17 | Balance trackers remain fundable after all corresponding mech factories are disabled | BalanceTrackerBase, BalanceTrackerFixedPriceToken, BalanceTrackerFixedPriceNative | Informative |
 
 The present document aims to point out some vulnerabilities in the autonolas-marketplace
 contracts.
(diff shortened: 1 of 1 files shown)


77ph (changes requested):
**Review — requesting one factual correction; the rest of entry #17 is accurate and well-reasoned.**

The mechanism analysis is right in most respects, but the entry's central reassurance is **factually incorrect** and, as written, inverts the actual risk.

**The error.** Entry #17 states: *"deposits remain withdrawable by the depositor at any time."* This is not true — **there is no requester-facing withdraw path in any balance tracker.**
- `grep -rn "function withdraw" contracts/` returns nothing. `BalanceTrackerBase` exposes only marketplace-gated accounting (`checkAndRecordDeliveryRates` / `finalizeDeliveryRates` / `adjustMechRequesterBalances`), mech payout (`processPayment[ByMultisig]`), and `drain()` (which moves only `collectedFees`). The internal `_withdraw(account, amount)` is called **only** from `_processPayment` to pay the **mech** — never to return `mapRequesterBalances[requester]` to the requester. `mapRequesterBalances` is only ever debited (payment) or credited (overpayment refund into the mapping); it is never paid back out as tokens.
- This directly **contradicts the repo's own vulnerability #7** ("No refund for expired undelivered requests", `BalanceTrackerBase`): *"there is no mechanism for the requester to reclaim the locked funds."*

**Consequence — the risk is the opposite of what the entry says.** On a retired tracker, a deposit can be neither processed as payment (no request path — correct) **nor** reclaimed by the depositor (no withdraw path). So the funds are **permanently stuck**, not "withdrawable." Framing them as retrievable + "no change recommended" makes a fund-lock look benign, and would lead off-chain monitoring to treat a non-zero retired-tracker balance as safe/recoverable when it is not.

**Requested fix (doc-only).**
1. Replace *"deposits remain withdrawable by the depositor at any time"* with the truth: deposits to a retired tracker are **stuck** — no payment path **and** no requester-reclaim path (cross-reference **vuln #7**). This makes #17 a corollary of #7 for retired branches (and it is **self-inflicted** — a voluntary deposit to a dead branch — which is worth stating).
2. In the trustlessness note, separate the two properties that are currently conflated: "no owner/governance/upgrade can **seize** the deposit" (true) is **not** the same as "the depositor can **retrieve** it" (false, per #7). Keep the former; drop any implication of the latter.
3. Adjust the closing monitoring guidance to: a non-zero balance on a retired tracker is **stuck funds**, not payment activity.

**Confirmed correct (no change needed):** the routing claim is properly hedged ("if no such mech was ever created…"); `drain()` moving only `collectedFees` is verified; and immutability meaning no party can seize/freeze/redirect deposits is verified. The single issue is the "withdrawable" claim.

77ph (approved):
Verified `f884fba` against source — the fix is correct and complete. Approving.

The corrected entry #17 is now factually accurate on all three points:
1. **"withdrawable" → "permanently stuck"** — grounded exactly as it should be: the concrete trackers expose no requester-facing withdraw function, and the internal `_withdraw()` is invoked only by `_processPayment()` to pay mechs, so `mapRequesterBalances` is never paid back out. Correctly framed as a corollary of vulnerability #7 for retired branches, and correctly flagged as self-inflicted (a voluntary deposit to a branch that can never serve a request).
2. **Non-seizability vs. retrievability** are now separated — no owner/governance/upgrade can seize/freeze/redirect deposits (kept), but the depositor cannot retrieve them either (cross-referenced to #7).
3. **Monitoring guidance flipped** — a non-zero balance on a retired tracker now reads as *stuck funds*, not payment activity.

No further changes needed. Thanks for the quick, precise fix.

